PR #1, opened 2023-03-01
Add: parse contributors script to update contribs

lwasser:
A start at processing contributor files and creating and updated YAML file with the info for each from github.
This means that we don't have to maintain user information we can just update from their github accounts!! 

## TODO

- [x] Add a website check that makes sure 1) the website starts with https (in some cases there is not http anything) and that it's https not http. this should be an easi string append / replace check on each contrib that happens AFTEr the github API update  
- [x] Add a helper that checks to see if a website works if it doesn't return a 200 then don't publish it on the website. 

```python
import requests

def check_url(url):
    try:
        response = requests.get(url)
        return response.status_code == 200
    except:
        return False

```


lwasser:
@xmnlab here is my first stab at a workflow for parsing contributors. you should be able to run it once you generate a API token locally to use. i'd love feedback on this ... i may turn it into a small module at some point. i'm also working on a parse issues script that could be another object maybe?

parse_contributors_oo.py
```diff
@@ -0,0 +1,313 @@
...
+                entry["website"] = entry.pop("profile")
+
+                # Process github image avatar id
+                entry["avatar_url"] = int(
+                    entry["avatar_url"].rsplit("/", 1)[-1].rsplit("?", 1)[0]
+                )
+                entry["github_image_id"] = entry.pop("avatar_url")
+                # Add empty values for the new keys
+                for key in [
+                    "mastodon",
```
lwasser, line 50: keys probably shouldn't be hard coded into this method. it is also long because of that

parse_contributors_oo.py
```diff
@@ -0,0 +1,313 @@
...
+# Get contribs from website
+web_contrib_path = "https://raw.githubusercontent.com/pyOpenSci/pyopensci.github.io/main/_data/contributors.yml"
+
+
+process_contribs = processContributors(json_files, API_TOKEN)
+# Combine the cross-repo contribut data
+all_contribs_dict = process_contribs.combine_json_data()
+
+# Open the web contrib file (could also be a method)
+with urllib.request.urlopen(web_contrib_path) as f:
```
lwasser, line 265: this could be a small helper and the web contrib path could be an attributed added to the class on instantiation?

parse_contributors_oo.py
```diff
@@ -0,0 +1,313 @@
...
+    web_contrib_dict = ruamel.yaml.safe_load(f)
+
+# Create a list of all gh usernames contributors from the website YAML
+all_web_contribs = []
```
lwasser, line 269: this should be a method too

parse_contributors_oo.py
```diff
@@ -0,0 +1,313 @@
...
+        all_web_contribs.append(item["github_username"].lower())
+    except:
+        # Check if there is a missing gh username in our website listing
+        print(item["name"], "is missing a github username")
+
+# TODO: stopped work here!!
+
+# Combine the web contrib with the cross-website contribs
+# making sure the web content is first
+# TODO: turn into function
```
lwasser, line 281: turn into method

parse_contributors_oo.py
```diff
@@ -0,0 +1,313 @@
+import json
+import ruamel.yaml
+import urllib.request
+import requests
+
+
+class processContributors:
```
xmnlab, line 7:
```suggestion
class ProcessContributors:
```

https://peps.python.org/pep-0008/#class-names

parse_contributors_oo.py
```diff
@@ -0,0 +1,313 @@
...
+        -------
+            A list of updated user data via a list of github usernames
+        """
+        all_user_info = {}
+        for gh_user in gh_usernames:
+            print("Getting data for: ", gh_user)
+            all_user_info[gh_user] = self.get_user_info(gh_user, API_TOKEN)
+        return all_user_info
+
+    def update_contrib_data(self, contrib_data, gh_data, update_keys):
```
xmnlab, line 158: missing the return type annotation

parse_contributors_oo.py
```diff
@@ -0,0 +1,313 @@
...
+            else:
+                cleaned_lines.append(line)
+
+        cleaned_text = "".join(cleaned_lines).replace("''", "")
+
+        with open(filename, "w") as f:
+            f.write(cleaned_text)
+
+
+"""Begin actual script """
```
xmnlab, line 247: not sure if you really need this code in the global scope of the module, 
but normally we create a function called `main` and we add everything there
and we call it at the end of the file with:

```
if __name__ = "__main__":
    main()
```

parse_contributors_oo.py
xmnlab: that makes sense to me

parse_issues.py
```diff
@@ -0,0 +1,76 @@
+import requests
```
xmnlab, line 1: not sure if there is a convention formally .. but usually we have just some global variables in the global scope of the module and the code actually we have inside functions .. maybe a `main` function and you can call it at the end of the file, as mentioned before.

xmnlab (commented):
Hi @lwasser I just gave a very superficial review here ..
if you want we can have a call to discuss more about this PR
looking forward to seeing your progress with OOP :)

lwasser:
@all-contributors please add @xmnlab for code, review, design

Issue #6, opened 2023-04-19
✨ Update package categories format in pyosmet update workflow to produce a better packages.yml file [Python programming]


lwasser:
Currently, the workflow in this repo will create an output file [called packages.yml that is a data file](https://github.com/pyOpenSci/pyopensci.github.io/blob/main/_data/packages.yml). 

[This data file drives our websites packages listing here. ](https://www.pyopensci.org/python-packages.html). The data file template for a single package is below. 

Output packages.yml file currently looks something like this (NOTE that i've edited the categories format manually - a separate issue to be opened):

Note in the yaml file below the categories often have two works with a space in between `Data munging` and sometimes there are capital letters. However for the website we need the categories to be lower case and using dashes rather than spaces. `data-munging`. I've been updating this manually but we should just update the code to fix the category style returned in the packages.yml file. 

```
- package_name: xclim
  package_description: Climate indices computation package based on Xarray
  submitting_author:
      name: Trevor James Smith
      github_username: Zeitsperre
  repository_link: https://github.com/Ouranosinc/xclim
  version_submitted: v0.40.0
  categories:
    - Data munging
    - Reproducibility
    - Geospatial
```

bbulpett:
✋ I would like to work on this one.

lwasser:
closing this now in favor of #20

PR #11, opened 2023-04-22
Fixed authentication to GitHub API.

austinlg96:
I was exploring some of the issues that you had mentioned may be good for the sprint at PyCon today and decided to se what I could do about #6.

I was running into the API limits described in some of the comments.

I believe this is because all of the github api requests are unauthenticated. [One call ](https://github.com/pyOpenSci/update-web-metadata/blob/10bb274b3eefbcb20676044ce25f9dd33988c3a6/src/pyosmeta/parse_issues.py#L49) seems to try and authenticate, but I don't think that the method used is actually supported by Github. The type of authentication attempted also explains [why the username was hardcoded](https://github.com/pyOpenSci/update-web-metadata/blob/10bb274b3eefbcb20676044ce25f9dd33988c3a6/parse_review_issues.py#L35).

Updating each of the `.get()` calls to use the `Authorization` header seemed to resolve the rate limit issues that I was running into while testing.

I also updated the one case where `HTTP 403` was being handled to 1) have a bit more helpful output and 2) actually `exit()` when it says that it will. This only applies in the one case though, so maybe calls to `requests.get()` could be replaced with a new function which makes the request and then can handle any HTTP errors in one place. Authentication may also be able to happen there too.

I removed the `try`\`except` block from the `_get_response` method because it seems to try and catch API errors but the `403`s don't throw a python error. The `response` var will also always be undefined. Overall it seems to me that we would want python errors to propagate upwards rather than be caught. If you'd prefer to keep the block, `except Exception:` would probably be better (though not as good as catching some other specific exception). I believe that `except` alone will catch things like `^C`. 

I'll be sure to stop-in to the sprint, but will be a bit late. Don't hold #6 for me if someone else wanted to work on it.

### DIFF
--- a/parse_review_issues.py
+++ b/parse_review_issues.py
@@ -32,8 +32,7 @@ issueProcess = ProcessIssues(
 )
 
 # Get all issues for approved packages
-# TODO: why is this method using my username??
-issues = issueProcess.return_response("lwasser")
+issues = issueProcess.return_response()
 # breakpoint()
 # Fixed:
 review = issueProcess.parse_issue_header(issues, 12)
--- a/src/pyosmeta/parse_issues.py
+++ b/src/pyosmeta/parse_issues.py
@@ -40,17 +40,16 @@ class ProcessIssues(YamlIO):
 
     # Set up the API endpoint
     # TODO: can i call the API endpoint above
-    def _get_response(self, username):
+    def _get_response(self):
         """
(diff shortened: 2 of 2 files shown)


lwasser:
@all-contributors please add @austinlg96 for code, review, design

lwasser:
@austinlg96 thank you for much for this pr!! working great locally. merging it!

PR #19, opened 2023-04-23
Update ProcessIssue.get_categories()

austinlg96:
Includes changes from #11.

Adds a `fmt = True` default option to `ProcessIsue.get_catgories()` which formats the categories as described in #6.

Also expands upon the original type-hinting to satisfy Pyright's `strict` mode.

Added `fmt` to the docstring and re-wrote the `issues` docstring to try and be more explicit that it is `list[list[str]]` rather than `list[str]`.

### DIFF
--- a/parse_review_issues.py
+++ b/parse_review_issues.py
@@ -32,8 +32,7 @@ issueProcess = ProcessIssues(
 )
 
 # Get all issues for approved packages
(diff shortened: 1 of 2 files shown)


lwasser:
thank you so much for this pr, Austin! it is working great. merging.

PR #38, opened 2023-07-05
Numerous bug fixes and update reviewers script

lwasser:
This is a draft PR with ugly code that starts the process of updating review information and adding reviewers / editors to our peer review contrib list

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ Returns
 - In some cases users haven't updated their name on GitHub. Their profile may either contain a first name or no name. In those instances we may update a name manually. If a
   name exists in the contributors.yml file and it has at least two words (first and last), we should leave it as is in the `contributors.yml` file.
 
-### `python3 parse_review_issues.py.py`
+### `python3 parse_review_issues.py`
 
 To run:
 `python3 parse_review_issues.py`
@@ -135,6 +135,13 @@ python3 parse_review_issues.py
 Create environment:
 
 `mamba env create -f environment.yml`
+
+## Notes
+
+First run parse-contributors.py - this will parse through and create a current up-to-date contributors.yml file.
+
+- Next run `parse_review_issues.py` - this will parse through all of our open issues and will create a new packages.yml file
+- Once you have that you can run update-reviewers: This not yet created script will take the output of parse-contributors (which can be ingested as a dict from the output pickle file) with a key being the gh username.
 
 ## Contributors ✨
 
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ Returns
 - In some cases users haven't updated their name on GitHub. Their profile may either contain a first name or no name. In those instances we may update a name manually. If a
   name exists in the contributors.yml file and it has at least two words (first and last), we should leave it as is in the `contributors.yml` file.
 
-### `python3 parse_review_issues.py`
+### parse_review_issues script
 
 To run:
 `python3 parse_review_issues.py`
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 .DS_Store
-contributors.yml
-packages.yml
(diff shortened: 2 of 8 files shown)


update-reviewers.py
```diff
@@ -0,0 +1,200 @@
...
+    ],
+}
+
+
+missing_contribs = []
+for pkgName, issue_meta in packages.items():
+    print(pkgName)
+    # Get review meta -
+    # start with reviewers and editor
+    for issue_review_role in ppl_types:
```
lwasser, line 65: This section has many conditionals because sometimes the field is empty, other times it has existing values and i need to check whether the value exists before adding it so there aren't duplicates. FINALLY sometimes the user doesn't exist so i need to add them if that is the case

update-reviewers.py
```diff
@@ -0,0 +1,200 @@
...
+        if issue_meta[issue_review_role]["name"] == "":
+            packages[pkgName][issue_review_role]["name"] = contribs[gh_user][
+                "name"
+            ]
+
+        # TODO - current issue is packages-reviewed and editor are being duplicated
+        # some of it might be case sensitive but...
+        # TODO - david's contrib types are not correct either he's missing submitting author and maintainer
+        # Update the contrib file with review package name
+        try:
```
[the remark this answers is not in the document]
lwasser: but it feels brittle and i want to ultimately make a function(s) but want the code to run first.

update-reviewers.py
```diff
@@ -0,0 +1,200 @@
...
+
+        # Finally check that peer-review is in the list of types
+        if "peer-review" not in contribs[gh_user]["contributor_type"]:
+            contribs[gh_user]["contributor_type"].append("peer-review")
+
+    # Now parse through all_current_maintainers
+    # Note that the first maintainer may be the submitting author
+    # TODO - is there a way to combine this with the above workflow
+    # the difference is there is a list of maintainers
+    print("Now on to maintainers")
```
lwasser, line 114: THe only reason that this is in a different section (the functionality is the same) is that i may have several maintainers in a list. so the input format is different than the format for a reviewer_1 entry that has a single reviewer and github user name. but it seems to me i could combine all of this code potentially.

update-reviewers.py
```diff
@@ -0,0 +1,153 @@
...
+                        gh_user,
+                        "No prob! I'll add them later",
+                    )
+                    missing_contribs.append(gh_user)
+    else:
+        print("All maintainers is missing in the review for ", pkgName)
+
+# Export to yaml
+process_contribs.clean_export_yml(contribs, "contribs.yml")
+process_contribs.clean_export_yml(packages, "packs.yml")
```
lwasser, line 144: there are issues with the YAML exporting with characters *id001 and &id001 i'm starting to wonder if that is ascii now or something?

lwasser: this has been fixed locally

lwasser:
This pr is getting big so i may merge and then work on cleaning up update_reviewers separately. here i do the following

1. merge new update reviewers code with the work mike did on the CLI script
2. organize TODO's so it's clear what needs to happen next
3. clean up extra print statements
4. fix a bug if a package url has a / at the end when creating the repo endpoint
5. several other cleanup steps and typing

README.md
```diff
@@ -59,7 +59,7 @@ Returns
 - In some cases users haven't updated their name on GitHub. Their profile may either contain a first name or no name. In those instances we may update a name manually. If a
   name exists in the contributors.yml file and it has at least two words (first and last), we should leave it as is in the `contributors.yml` file.
 
-### `python3 parse_review_issues.py.py`
+### `python3 parse_review_issues.py`
```
lwasser, line 62:
```suggestion
### parse_review_issues script
```

PR #40, opened 2023-07-15
reusable workflow for calling scripts

msarahan:
This simplifies clients that are calling our scripts here, such as https://github.com/pyOpenSci/pyopensci.github.io/blob/main/.github/workflows/update-packages.yml

It is in response to the path being wrong with a recent effort to update packages, https://github.com/pyOpenSci/pyopensci.github.io/actions/runs/5562999146/jobs/10161656266#step:5:69

```
Notice:  To update, run: pip install --upgrade pip
/home/runner/work/pyopensci.github.io/pyopensci.github.io
python: can't open file '/home/runner/work/pyopensci.github.io/pyopensci.github.io/scripts/parse_issue_metadata.py': [Errno 2] No such file or directory
```

I'm mixed on whether this is the "best" approach. It's nice that it separates the python environment stuff over here, away from the repos that are just web content. At the same time, it pushes us further away from a unified local/remote development workflow.

A different approach is to not use github reusable workflows, and to reorganize this repo a bit to make the scripts be more entry-point friendly, such that after running `pip install`, we would have them accessible by name instead of by path.

This PR changes API_TOKEN to GITHUB_TOKEN to reflect the fact that github actions provides GITHUB_TOKEN for you automatically as a repo-scoped token. It's better practice than a stored secret of a user PAT.

### DIFF
--- a/.github/workflows/run-script.yml
+++ b/.github/workflows/run-script.yml
@@ -1,0 +1,28 @@
+on:
+  workflow_call:
+    inputs:
+      script_name_with_args:
+        required: true
+        type: string
+
+jobs:
+  run_script_job:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out the code
+        uses: actions/checkout@v3
+      - name: Setup Python
+        uses: actions/setup-python@v4
+      - name: Upgrade pip
+        run: |
+          # install pip=>20.1 to use "pip cache dir"
+          python -m pip install --upgrade pip wheel
+      - name: Install deps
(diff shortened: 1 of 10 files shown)


lwasser:
@msarahan just thinking about this more. just following up on a few items here
I didn't realize that GITHUB_TOKEN could be repo scoped. so that is perfect. 

Regarding making the scripts entry-point friendly, would that entail moving more of that workflow into a function / class within the pyos-meta package?  Many thanks. it's just helpful for me to understand better practices here !

PR #41, opened 2023-07-16
move scripts into src/pyosmeta/cli folder

msarahan:
This alters the scripts a tiny bit, making them more like typical CLI scripts, which would have a main() function, and then a separate bit at the end to parse arguments. This structure makes CLI scripts more testable, and it's also how you specify console scripts to PDM. The work here creates small shim "programs" that launch the scripts. These programs are not necessarily named the same as the script files. I chose to make them hyphenated instead of underscored, because hyphens are faster to type.

The implication here is that instead of copying scripts around, you can `pip install pyosmeta` and be left with CLI commands to run the scripts, instead of trying to figure out where they are and then calling python on them to run them. The CLI commands are on PATH when the environment is active, so you can call these commands no matter what your CWD is.


msarahan:
@lwasser this should fix the failing workflow. It's kind of a chicken/egg thing to test, unfortunately. This will change your local workflow, so I definitely want you to understand it before it gets merged. The console scripts are nicer ways to have a CLI, but any change to an existing workflow is disruptive.

msarahan:
Here's a build that shows things working: https://github.com/msarahan/update-web-metadata/actions/runs/5565453450

Before this PR gets merged, we should change the install URL for update-web-metadata back.

.github/workflows/run-script.yml
```diff
@@ -9,20 +9,19 @@ jobs:
...
         uses: actions/setup-python@v4
       - name: Upgrade pip
         run: |
           # install pip=>20.1 to use "pip cache dir"
           python -m pip install --upgrade pip wheel
-      - name: Install deps
-        run: python -m pip install -r requirements.txt
       - name: Install package
-        run: python -m pip install -e .
+        run: python -m pip install git+https://github.com/msarahan/update-web-metadata.git@console-scripts
```
willingc, line 20: @msarahan Do you plan to merge with this pointing to your repo or are you just leaving for now for testing?

pyproject.toml
```diff
@@ -18,4 +18,9 @@ requires-python = ">=3.10"
 readme = "README.md"
 license = {text = "MIT"}
 
+[project.scripts]
+parse-contributors = "pyosmeta.cli.parse_contributors:main"
```
willingc, line 22: Nice :sunny:

willingc (approved):
LGTM @msarahan. I'm assuming that the deleted and added files are similar other than adding standard main() behavior.

.github/workflows/run-script.yml
lwasser: oh i do have a similar question as @willingc , Mike. Many thanks.

pyproject.toml
lwasser: oh wow - is this an entry point? i've read a bit about this but am learning here and REALLY appreciate that!! :)

.github/workflows/run-script.yml
```diff
@@ -9,20 +9,19 @@ jobs:
   run_script_job:
     runs-on: ubuntu-latest
     steps:
-      - name: Check out the code
-        uses: actions/checkout@v3
+      # TODO: consider replacing python/pip/update-web-metadata installs with docker image
```
lwasser, line 12: i can build a container. will that make the workflow faster?

msarahan: It's not going to get a ton faster, because it already doesn't take much time. It would save some bandwidth for pypi, but again not much. It's only worthwhile if it would also be useful for you in other ways.

.github/workflows/run-script.yml
msarahan: The plan is to revert the commit that starts with 'REVERT ME' before merging. I'm on the road away from a real computer right now. I might be able to handle that this evening. Please feel free to push to my fork/branch if you want to merge this sooner.

PR #42, opened 2023-07-20
Code refactor & numerous bug fixes 

lwasser:
well this is round two of trying to do better. DRY == good


src/pyosmeta/parse_issues.py
```diff
@@ -84,6 +84,15 @@ def _contains_keyword(self, string: str) -> bool:
             ("Submitting", "Editor", "Reviewer", "All current maintainers")
         )
 
+    def _clean_name(self, a_str: str) -> str:
+        """Helper to strip unwanted chars from text"""
+
+        unwanted = ["(", ")", "@"]
```
lwasser, line 90: could make this ore flexible by making unwanted a var with a default list value

src/pyosmeta/cli/update_contributors.py
```diff
@@ -0,0 +1,79 @@
...
+                web_contribs.update(
+                    processContribs.check_add_user(gh_user, web_contribs)
+                )
+
+            # Update contrib type list
+            existing_contribs = web_contribs[gh_user]["contributor_type"]
+            web_contribs[gh_user][
+                "contributor_type"
+            ] = processContribs.update_contrib_list(existing_contribs, key)
+
```
lwasser, line 51: TODO: would like to add an update flag that is optional for this workflow.

src/pyosmeta/cli/update_review_contribs.py
```diff
@@ -0,0 +1,143 @@
...
+                        contrib_key_yml = contrib_types[issue_review_role][0]
+                        contrib_key = contribs[gh_user][contrib_key_yml]
+                        # Update contrib roles
+                        review_role = contrib_types[issue_review_role][1]
+
+                        # Update contrib packages for peer review
+                        contribs[gh_user][
+                            contrib_key_yml
+                        ] = updateContribs.update_contrib_list(contrib_key, pkg_name)
+
```
lwasser, line 83: Question: how to use aliases so they update a dictionary or should i use them this way?

how do alias' behave within functions??

src/pyosmeta/cli/update_review_contribs.py
```diff
@@ -0,0 +1,143 @@
...
+        print("Processing review team for:", pkg_name)
+        for issue_review_role in contrib_types.keys():
+            if issue_review_role == "all_current_maintainers":
+                if issue_review_role in issue_meta:
+                    # Loop through each maintainer in the list
+                    for i, a_maintainer in enumerate(issue_meta.get(issue_review_role)):
+                        gh_user = get_clean_user(a_maintainer["github_username"])
+
+                        # TODO: The steps below are almost identical to the reviewers
+                        # Could these alias' be defined only once?
```
lwasser, line 66: TODO: this enture section could be a method that is used here for maintainers and below for edits/reviewers etc.
but hung up on how to do this if we use aliases or what the best way is

src/pyosmeta/contrib_review_meta.py
```diff
@@ -3,26 +3,40 @@
...
 
 @dataclass
 class UpdateReviewMeta(ProcessContributors):
-    def __init__(self):
+    """This is a class that is designed to add additional functionality
+    to ProcessContribs to support updating reviewer metadata associated with
+    review issues. It is a bridge between the ProcessContributors and
+    ProcessReviews objects."""
+
+    def __init__(self, GITHUB_TOKEN):
```
lwasser, line 18: if i inherit from another class what is the best way to inherit attrs such as the token?

lwasser:
@xmnlab if you have any time next week would you please review this for me. I'd love to know if i could make this more efficient / cleaner etc. many many thanks!!

src/pyosmeta/cli/update_contributors.py
```diff
@@ -0,0 +1,77 @@
...
+# that are used in a method that i call via composition are the helpers
+# still available?
+
+
+def main():
+    GITHUB_TOKEN = get_api_token()
+    update_all = False
+
+    # TODO - maybe add these as an attr in the contribs class?
+    json_files = [
```
xmnlab, line 19: maybe would could have a variable to store `https://raw.githubusercontent.com/pyOpenSci` so you don't need to repeat it everytime

src/pyosmeta/cli/update_contributors.py
```diff
@@ -0,0 +1,77 @@
...
+        "https://raw.githubusercontent.com/pyOpenSci/software-peer-review/main/.all-contributorsrc",
+        "https://raw.githubusercontent.com/pyOpenSci/pyopensci.github.io/main/.all-contributorsrc",
+        "https://raw.githubusercontent.com/pyOpenSci/software-review/main/.all-contributorsrc",
+        "https://raw.githubusercontent.com/pyOpenSci/update-web-metadata/main/.all-contributorsrc",
+    ]
+
+    # Get existing contribs from pyopensci.github.io repo (website data)
+    web_yaml_path = "https://raw.githubusercontent.com/pyOpenSci/pyopensci.github.io/main/_data/contributors.yml"
+
+    processContribs = ProcessContributors(json_files, GITHUB_TOKEN)
```
xmnlab, line 30:
```suggestion
    process_contribs = ProcessContributors(json_files, GITHUB_TOKEN)
```

src/pyosmeta/file_io.py
```diff
@@ -21,124 +20,158 @@ def get_api_token() -> str:
...
+    """
+    final_dict = {}
+    for dict in a_list:
+        # All github usernames are lower to make this a key
+        final_dict[dict[a_key].lower()] = dict
+
+    return final_dict
+
+
+def load_website_yml(a_key: str, a_url: str):
```
xmnlab, line 49:
```suggestion
def load_website_yml(key: str, url: str):
```

it is not common to have an `article` as part of the variable name

xmnlab: also, if it is possible, annotate the type for the output, for example:

`def load_website_yml(key: str, url: str) -> Dict[THE_KEY_TYPE, THE_VALUE_TYPE]:`

src/pyosmeta/file_io.py
```diff
@@ -21,124 +20,158 @@ def get_api_token() -> str:
...
-    A class that provides file output and cleanup methods.
+    This opens a website contrib yaml file and turns it in a
+    dictionary
+    """
+    yml_list = open_yml_file(a_url)
+
+    return _list_to_dict(yml_list, a_key)
+
+
+def dict_to_list(pyosDict: dict) -> list:
```
xmnlab, line 59: instad of `pyosDict` (cammel case), consider using snake case. also maybe it would be nice something more descriptive, but maybe it is a very generic dictionary, so maybe you can use something like `metadata`, `github_data`, etc. Normally it is not a good approach to have the type of object as part of the variable name ... but I guess that if you have the same kind of data in a different type, for example a date in string format and the same date in datetime type, it makes sense to me to have the type as part of the variable name, but I guess that in this case it is not necessary.

xmnlab: also, maybe you could improve the type annotation to add more information, for example,
instead of `list`, you could have `List[str]` to represent a list of strings.

but, maybe this function is not super necessary as python already have a built-in function to get the value from a dictionary:

```
>>> {"a": 1, "b": 2, "c": 3}.values()
dict_values([1, 2, 3])
```

src/pyosmeta/cli/update_review_contribs.py
```diff
@@ -0,0 +1,137 @@
...
+from pyosmeta.contributors import ProcessContributors
+from pyosmeta.file_io import clean_export_yml, get_api_token, load_pickle
+
+
+def get_clean_user(username: str):
+    return username.lower().strip()
+
+
+def main():
+    GITHUB_TOKEN = get_api_token()
```
xmnlab, line 39: if you are not planning to have different kind of connections
maybe you could move this step directly to ProcessContributors.
in another words, `ProcessContributors` could get the api token directly

src/pyosmeta/file_io.py
```diff
@@ -21,124 +20,158 @@ def get_api_token() -> str:
...
+
+    a_list : list
+        A list of dictionary objects returned from load website yaml
+    a_key : str
+        A string representing the dict key to use as the main key to call
+        each sub dict in the object.
+
+    """
+    final_dict = {}
+    for dict in a_list:
```
xmnlab, line 42: `dict` is a reserved word, so normally it is not a good practice to use that name.
also maybe you could use list comprehension here, something like:

`return { data[a_key].lower(): data for data in a_list }`

xmnlab (commented):
Hi @lwasser ! just give a superficial review here, in general my suggestions would be:

* avoid using camelCase naming style in Python
* you don't need to use `a_` to prefix a variable, normally we don't use an article as prefix for the names
* for typing annotation, specifically for dictionaries and lists, consider using `Dict` instead of `dict` and `List` instead of `list`, also, add the type of the items, for example `List[str]` or `Dict[str, List[str]]` so it would be much for informative what kind object the function will receive as a parameter or what kind of object it will return.
* add the typing annotation for the function output (returning type)
* when you have some strings that are repeated too much in the code, consider using a variable to store that string.

you are doing a great job! looking forward to seeing your progress here :)

src/pyosmeta/cli/update_contributors.py
lwasser: fixed!  i also made a small list comprehension!

```
base_url = "https://raw.githubusercontent.com/pyOpenSci/"
    end_url = "/main/.all-contributorsrc"
    repos = [
        "python-package-guide",
        "software-peer-review",
        "pyopensci.github.io",
        "software-review",
        "update-web-metadata",
    ]
    json_files = [base_url + repo + end_url for repo in repos]
```

src/pyosmeta/cli/update_review_contribs.py
lwasser: that worked!

src/pyosmeta/file_io.py
lwasser: that's much cleaner. i always forget about comprehensions for dictionaries and i'm not sure why.

src/pyosmeta/file_io.py
lwasser: i'm curious about this one. for me it just makes it easier (for me) to read / understand. particularly because keys() is a method. is this a python pep type standard or just not a normal way that people create these types of variables? i can see how a_url might be unnecessary as it's not related to python and explicit enough.

src/pyosmeta/file_io.py
lwasser: i've fixed most of this but for some reason when i call `.values()` on the Dict i get a representation error when i try to export my data to YAML. i still haven't figured out why.

lwasser: chatgpt for the win

```
>>> list({"a": 1, "b": 2, "c": 3}.values())
[1, 2, 3]
```

if you don't make it a list then it just returns a dict_values() thing that is not a list!

lwasser: some of this stuff i just haven't had time to dig into and refine. but i still have design questions generally. but will make these fixes for now as it's cleaner!!

lwasser:
Thank you @xmnlab i am going to merge this as things are actually working now!! But i'll spend more time on refactoring, typing and cleanup as i continue to develop it further! i did figure out how to call that api token using composition and it is working really well.  🎉

xmnlab:
@lwasser , awesome! really glad that it worked fine! congrats! :tada:

PR #55, opened 2023-08-14
Refactor: move to pydantic for validation

lwasser:
pydantic is so cool! but i can't figure out how to update (parts of) a model and apply validation while doing so.


lwasser:
yay - all seems to be working as expected. doing to try to force run in the website repo to see

PR #56, opened 2023-08-17
Update pyproject.toml

lwasser:
This is a test for using pdm and an updated pyproject.toml file which is the current plan for our tutorials.


lwasser:
This is failing because of flake8 which i fixed in this pr #55 but that pr is not ready to merge yet - so i'm going to merge this failing as it's package infrastructure.

Issue #70, opened 2023-10-15
Error parsing user metadata


SimonMolinsky:
# Problem

There is a problem with parsing contributors' metadata in package reviews. The problem is probably related to [this line of the code](https://github.com/pyOpenSci/update-web-metadata/blob/790de74fc54c4da2e7cefde0bd8533055bd8859f/src/pyosmeta/parse_issues.py#L311C17-L311C51):

```python
    def _get_line_meta(self, line_item: list[str]) -> dict[str, object]:
        ...
        if self._contains_keyword(line_item[0]):
            if line_item[0].startswith("All current maintainers"):
                ...
            else:
                names = line_item[1].split("(", 1)  # THIS LINE HERE!
               ...
        return meta
```
It probably leads to the parsing error here: https://github.com/pyOpenSci/pyopensci.github.io/actions/runs/6521128026/job/17709405005

# Reason

I think splitting the line by round bracket `)` is why the error occurs. One reviewer's name here: https://github.com/pyOpenSci/software-submission/issues/115 has this character. 

# Solution

There are two options:

1. Restrict characters in names (update Editors Guide),
2. Change the parsing method to detect the `@` character.

**I can do both, so you can assign me to this issue** :rocket:

cmarmo:
@SimonMolinsky I'm having the same issue when updating the website for astartes. It would be wonderful if you can fix it!
As far as I'm concerned I'd rather have the `@` detected.

lwasser:
@SimonMolinsky if you can work on this i'd greatly appreciate it!! i must have broken the build with a recent update!!  i'll assign you now.

lwasser:
also agree with @cmarmo  on - let's parse for the @. i haven't look at that code in a bit of time but i DO recall i used the ghusername as a unique identifier in the dictionaries (i'm pretty sure each dict contains a pydantic "personmodel" that i created and validates accordingly. But it is the case that sometimes in our issues, people add their names in ways that i didn't expect. then it breaks the build. 

i'm wondering if that is the case here. but regardless appreciate your looking into this - thank you!!

SimonMolinsky:
@lwasser It was relatively easy to dive into your code. It is self-explaining. I will work on it then. Thanks for the assignment!

@cmarmo I will check your case too :)

lwasser:
[...]

PR #71, opened 2023-10-20
Update parse usernames process

SimonMolinsky:
#70 

I've created a function outside `ProcessIssues()` class for parsing usernames. Additionally, I've written a simple unit test. I didn't check the process of updating webpage data because I don't have a Github Token for it (**maybe we need some toy repository to test this package?**).

🐢

### DIFF
--- a/src/pyosmeta/parse_issues.py
+++ b/src/pyosmeta/parse_issues.py
@@ -37,6 +37,56 @@ def clean_date(a_date: Optional[str]) -> str:
             return "missing"
 
 
+def clean_name(a_str: str) -> str:
+    """Helper to strip unwanted chars from text"""
+
+    unwanted = ["(", ")", "@"]
+    for char in unwanted:
+        a_str = a_str.replace(char, "")
+
+    return a_str.strip()
+
+
+def parse_user_names(username: str) -> dict:
+    """Parses authors, contributors, editors and usernames from requested issues.
+
+    Parameters
+    ----------
+    username : str
+        The line with username (see Notes).
+
+    Returns
+    -------
+    : dict
+        ``{name: str, github_username: str}``
+
+    Notes
+    -----
+    Possible combinations:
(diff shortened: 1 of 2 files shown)


lwasser:
@SimonMolinsky thank you for this - i'll have a look at it locally now!! 

and yay we have our first test !! i've wanted to add tests i just haven't had time. thank you !!!
as far as a test repository. we can totally have something like that. i guess we need to think about whether it's needed. in the other repo i think the token is for opening the PR. BUT perhaps you are right there is an api token for accessing github.

you should be able to run this all locally - so i'd love to understand what didn't run or what was confusing about the workflow so i can add documentation. so far it's only been me working on it. Having fresh eyes is really useful.

lwasser:
amazing! it runs locally. i'm going to push a small enhancement that then validates package names using the same cleanup step you used for names. i noticed astartes has markdown in the name!! but let's merge this now and then we can move forward with that other cleanup step!! 

@SimonMolinsky i've always used pytest to run tests. in this case how would we run the tests? 
many thanks

lwasser:
@all-contributors please add @SimonMolinsky for code, review

SimonMolinsky:
Two new things appeared in this PR, so I will start with the first:

> you should be able to run this all locally - so i'd love to understand what didn't run or what was confusing about the workflow so i can add documentation. so far it's only been me working on it. Having fresh eyes is really useful.

Ok, I didn't grasp it because I don't know what exactly is the output of each process, and if I can run processes separately. I will experiment with it and tell you what is/was confusing (or rather - missing) in README.

> @SimonMolinsky i've always used pytest to run tests. in this case how would we run the tests?

I've used `unittest` because I don't see `pytest` as a dependency. If you wish, I can change the code and add `pytest` to dependencies, but in another Issue I suppose :) I prefer `pytest` too :)

PR #90, opened 2024-02-21
Fix: move to hatch scripts for tests and move to pytest

### LINKED ISSUE #89: Setup test suite runs with pytest and hatch scripts
Hatch has built in scripts that allow for nox like functionality. It would be good to setup our test suite to support scripts rather than adding another tool to our repo. it would also be good to use pytest rather than unittest because we can take advantage of parametrize which is pretty awesome and will streamline our test complexity significantly.

lwasser:
closes #89 

This is a small PR to begin moving to hatch scripts for easier test runs and automation. 
i also moved us over to pytest to take advantage of pytest.parametrize

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -12,13 +12,7 @@ maintainers = [
 ]
 
 classifiers = [
-    # How mature is this project? Common values are
-    #   3 - Alpha
-    #   4 - Beta
-    #   5 - Production/Stable
     "Development Status :: 4 - Beta",
-
-    # Indicate who your project is intended for
     "Intended Audience :: Developers",
     "Topic :: Software Development :: Build Tools",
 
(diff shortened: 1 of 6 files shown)


pyproject.toml
```diff
@@ -71,10 +61,23 @@ update-contributors = "pyosmeta.cli.update_contributors:main"
 update-reviews = "pyosmeta.cli.process_reviews:main"
 update-review-teams = "pyosmeta.cli.update_review_teams:main"
 
+### Hatch config ###
+
 [tool.hatch]
 version.source = "vcs"
 build.hooks.vcs.version-file = "src/pyosmeta/_version.py"
 
+[tool.hatch.envs.test]
```
lwasser, line 70: This sets up a test envt with whatever deps you want

pyproject.toml
```diff
@@ -71,10 +61,23 @@ update-contributors = "pyosmeta.cli.update_contributors:main"
...
+dependencies = [
+  "pytest",
+  "pytest-cov",
+]
+
+[tool.hatch.envs.test.scripts]
```
lwasser, line 76: this sets up a script that like nox can be run. 

so running `hatch run test:run-tests`
where test is the envt name and run-tests is the script to call will run our tests and make sure pytest is installed. you can also setup envt defaults if you want

Issue #91, opened 2024-02-22
Refactor needed for parse_issue_header method


lwasser:
Phew - working on the code i'm noticing some VERY convoluted structures i have around parsing the issue metadata that make it very hard to follow the code. In the next PR where i manage partnership data, i have the todo below added to the code:   
```
 # TODO: this method parse_issue_header is working but the parsing code is 
    # really hard to follow
    # It is worth another pr that cleans up the workflow around grabbing 
    # metadata so it's clearer to follow.  
    accepted_reviews = process_review.parse_issue_header(issues, 45)
```

we need to refactor this part of the workflow so 

1. the methods are clearly named and scoped to do specific things and it's clear what happens when
2. the docstrings clarify what the methods do on top of the names of the methods being clear
3. I suspsect refactoring would also mean some code cleanup in general as the method now returns a bunch of extra unprocessed data that doesn't need to be returned. 

specifically in this part of the process_Reviews.py script

```python
final_reviews = {}
    for key, review in all_reviews.items():
        # First add gh meta to each dict
        print("Parsing & validating", key)
```

the review data returned looks like the image below. Note that in the image below, there is a bunch of raw github issue data that hasn't been cleaned 

like this line 
`'- [x] I have read and will commit to package maintenance after the review as per the [pyOpenSci Policies Guidelines][Commitment].'`

and then there is some nicely formatted data which is what we want that method to return like this:
```
`date-accepted`: '02/06/2024'
```
that message uncleaned data above shouldn't be returned at all. 

the code is working now because the pydantic model just ignores those elements. I think our methods should be more intentional about returning clean data. They shouldn't just work because it works.

<img width="1067" alt="Screenshot 2024-02-21 at 7 33 18 PM" src="https://github.com/pyOpenSci/pyosMeta/assets/7649194/b5b65470-c123-471f-8e11-b7e2a8f65b21">

PR #93, opened 2024-02-22
Add: partner support to metadata parsing

### LINKED ISSUE #92: Add: support for astropy and other partners in the parse issue method
Right now parse issues returns issue metadata. 
We haven't until now needed to parse partner data. 

I have updated the [sunpy review here](https://github.com/pyOpenSci/software-submission/issues/147) with the format that i'd like the new issue template to look like. 

We will want to parse that ## Community partners section and pull any partner information into a list that we can parse on our website (and that partners can parse as well). 

## Steps

- [ ] first add "get_partners" to the issue parse workflow & create tests for it working as expected in various cases
- [ ] Then add get_partners as a part of the contributor workflow to add partner data to contribs + tests

lwasser:
closes #92 

This is a small fix to parse the partners in our issues and then return the value in our model. 
NOTE: need to test the full workflow including exporting to yaml before merging this so i'll open as a draft. i suspect i have to add a key somewhere to reorder the data. 

Also note that this pr inspired #91 because the code to parse the header is super hard to follow.


src/pyosmeta/parse_issues.py
```diff
@@ -167,6 +167,7 @@ class ReviewModel(BaseModel):
     closed_at: Optional[str] = None
     issue_link: str = None
     joss: Optional[str] = None
+    partners: Optional[list[str]] = None
```
lwasser, line 170: i believe i'll need to add partners in the export keys as well to ensure it exports to yaml. 

then it would be good to add some tests around various potential "cases" of data that could be found in an issue around the partner keys

lwasser:
This is working now BUT the categories are doing something funky -- 
```
    - data-retrieval
    - data-extraction
    - data-processing/munging
    - data-deposition
    - data-validation and-testing
    - workflow-automation
    - database-interoperability
```

will revisit that piece and potential pydantic validators to cleanup.

PR #99, opened 2024-02-24
Remove unneeded file now that hatch is used for packaging

willingc:
Removes `__about__.py` that is no longer needed to set a version.


lwasser (approved):
Thank you @willingc good catch on this file not being needed anymore!!
i will merge this. in general are there rules around when to update the changelog vs not? i'm asking because this pr doesn't include a changelog but it's also very small. so i wasn't sure what the general thoughts are on this. i will still merge this i'm just wondering.

Issue #78, opened 2023-12-16
Update repo for maintainability


willingc:
Evolving code bases often need some "cleaning" to help the repo be understandable and maintainable by future contributors. This issue will track the series of steps to be taken while cleaning.

## Phase 1
Update repo root for maintainability:

- [x] Simplify README for readability
    - Add badges
    - Create sections that would be expected
    - Update title to reflect that the repo will be renamed in the future to the `pyosmeta` to match the package name
- [x] Create a `CONTRIBUTING.md` file and move usage instructions from existing README to it
- [x] Create a `CODE_OR_CONDUCT.md` file and add text to refer to the pyOpenSci CoC page
- [x] Create a `CHANGELOG.md` file which will contain a badge to PyPI for first iteration
- [x] Create a `LICENSE` file at the repo root for MIT license with content from `src/LICENSE.txt` and drop the `.txt` suffix
- [x] Remove `src/LICENSE.txt` and `src/README.md`

## Phase 2
- [x] Review the development tools config (flake8, recommit, gitignore, etc.)

## Phase 3
- [ ] Correct any errors surfaced by running code inspectors

## Phase 4
- [ ] Add module/file docstrings
- [ ] Review function, class, and public method docstrings and edit as needed

## Phase 5
Begin code enhancements such as adding a CLI that allows running: `pyosmeta <task> <commandline>`

willingc:
@lwasser We'll be pairing on this with you doing the bulk of the PR reviews.

lwasser:
@all-contributors please add @willingc for code, review

lwasser:
yes please @willingc 

excited to work on this with you! i also just added two other issues that will improve the package's infrastructure. i'll do a better job of updating the changelog as well from now on. because i've been the only person working on this, I know it's missing a lot of core package elements.  I appreciate getting to work with you on this!

willingc:
Closing this since I think we will cover Phase 3, 4, and 5 during incremental refactor of the repo.

PR #101, opened 2024-02-24
Update gitignore, flake8, and pre-commit config files

willingc:
Addresses #78.

This PR updates and documents the following configuration files:
- .gitignore (use standard Python gitignore and move pyosMeta specific items to end)
- .flake8 (explain ignored rules and add a code complexity check)
- .precommit.config.yaml (improve comments for maintainability and new contributor friendliness)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -166,19 +166,7 @@ cython_debug/
 .vscode/
 
 # pyosMeta specific
-all_contribs.pickle
-all_contribs_dict.pickle
-all-submissions.csv
-issue-stats.ipynb
-plot.py
-presubmissions.csv
-reviews_all.csv
-reviews_presubmissions.csv
 src/pyosmeta/_version.py
-src/pyosmeta/cli/test.ipynb
-
-contrib*.yml
-pack*.yml
 *.pickle
 
 .token
--- a/.gitignore
+++ b/.gitignore
@@ -1,25 +1,123 @@
-.DS_Store
-contrib*.yml
-pack*.yml
-*.pickle
-all_contribs.pickle
-all_contribs_dict.pickle
-.env
-.vscode/
-.idea/
-.token
-__pycache__
-.coverage
-coverage.xml
+# Byte-compiled / optimized / DLL files
+__pycache__/
+*.py[cod]
+*$py.class
 
+# C extensions
+*.so
 
 # Distribution / packaging
 .Python
 build/
+develop-eggs/
 dist/
+downloads/
+eggs/
+.eggs/
+lib/
+lib64/
+parts/
+sdist/
+var/
+wheels/
+share/python-wheels/
+*.egg-info/
+.installed.cfg
+*.egg
+MANIFEST
+
+# PyInstaller
+#  Usually these files are written by a python script from a template
+#  before PyInstaller builds the exe, so as to inject date/other infos into it.
+*.manifest
+*.spec
+
+# Installer logs
+pip-log.txt
+pip-delete-this-directory.txt
+
+# Unit test / coverage reports
+htmlcov/
+.tox/
+.nox/
+.coverage
+.coverage.*
+.cache
+nosetests.xml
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -1,26 +1,124 @@
...
+#   https://python-poetry.org/docs/basic-usage/#commit-your-poetrylock-file-to-version-control
+#poetry.lock
+
+# pdm
+#   Similar to Pipfile.lock, it is generally recommended to include pdm.lock in version control.
+#pdm.lock
+#   pdm stores project-wide configurations in .pdm.toml, but it is recommended to not include it
+#   in version control.
+#   https://pdm.fming.dev/#use-with-ide
+.pdm.toml
```
lwasser, line 110:
```suggestion

```
@willingc i had intentionally removed a lot from the .gitignore because we don't use a lot of this such as pdm. Do you think it's better practice to just include the stock .gitignore as is rather than being intentional about what would actually potentially be used / ignored in our workflow? i'm just curious. this definitely had some things i had to add back such and code coverage files. but then it has other things like this that we just don't need because we don't use it and would never have a pdm toml file in our build or dev workflow using hatch.

.gitignore
```diff
@@ -1,26 +1,124 @@
...
+
+# PEP 582; used by e.g. github.com/David-OConnor/pyflow and github.com/pdm-project/pdm
+__pypackages__/
```
lwasser, line 113:
```suggestion

```
similarly i think this pep isn't used? or that was my understanding. pdm supports it but i had heard it never was adopted (i could be wrong)

.gitignore
```diff
@@ -1,26 +1,124 @@
...
+
+# Celery stuff
+celerybeat-schedule
+celerybeat.pid
+
+# SageMath parsed files
+*.sage.py
```
lwasser, line 120:
```suggestion

```

.gitignore
```diff
@@ -30,19 +128,58 @@ ENV/
...
 
+# Spyder project settings
+.spyderproject
+.spyproject
+
+# Rope project settings
+.ropeproject
+
+# mkdocs documentation
+/site
```
lwasser, line 139:
```suggestion

```

we don't use mkdocs or rope

.gitignore
```diff
@@ -30,19 +128,58 @@ ENV/
...
 plot.py
 presubmissions.csv
 reviews_all.csv
 reviews_presubmissions.csv
+src/pyosmeta/_version.py
 src/pyosmeta/cli/test.ipynb
+
+contrib*.yml
+pack*.yml
+*.pickle
```
lwasser, line 182:
```suggestion
src/pyosmeta/_version.py
*.pickle
```

This is my mess 😄 i have learned to use a global .gitignore for dummy test files :)  so we can clean this up too as i have a better workflow now!

.pre-commit-config.yaml
```diff
@@ -1,55 +1,55 @@
...
-# - Register git hooks: pre-commit install --install-hooks
+# pre-commit (https://pre-commit.com/) is a tool that runs source code checks
+# such as linting, formatting, and code style.
+
+# CI
+# pyosMeta uses pre-commit with the precommit.ci bot to check pull requests.
+# Configuration reference doc: https://pre-commit.com/#pre-commit-configyaml---top-level
+
+# Developers
+# You may want to install pre-commit to run locally on each minimum_pre_commit_version
```
lwasser, line 9: Thank you this is wonderful!

lwasser (approved):
@willingc thank you for this. see what you think about removing some of the things we don't need or let me know if you think we should keep the default as is. once we clean this up we can merge this. approving!

.gitignore
willingc: I think it is personal preference. I tend to use what is the common standard that GH uses. Your call. I'm cool with whatever. I do like to have pyosMeta specific stuff at the end :smile:

PR #103, opened 2024-02-24
Update changelog to reflect actual history on PyPI and GitHub

willingc:


lwasser (approved):
Thank you for this, @willingc i hope we can  create a new release soonish to align pypi with GitHub! i don't know how that happened but i'd guess i was playing around with hatch and versioning and running tests and never fully cleaned things up! let's merge this

PR #104, opened 2024-02-24
Update optional dependencies and edit pyproject.toml

willingc:
This PR adds dev dependencies as optional dependencies and makes minor style changes.

Developers can now run `pip install ".[dev]"` or `pip install -e ".[dev]"` to install dev tools locally.

We can set up a dev environment for hatch in a separate PR.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -10,7 +10,6 @@ authors = [{ name = "Leah Wasser", email = "leah@pyopensci.org" }]
 maintainers = [
     { name = "pyOpenSci", email = "admin@pyopensci.org" }, # Optional
 ]
-
 classifiers = [
     "Development Status :: 4 - Beta",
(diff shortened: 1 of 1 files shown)


lwasser (approved):
Excellent. let's merge this. @willingc when we merge the tests PR, you'll notice i setup a hatch environment to run tests. We can actually setup a lint environment too or a general dev environment as well. 

this is where we'll learn that using hatch - there is a slightly different approach in terms of environments. i'm actually really enjoying it the more i use it. because you don't HAVE to worry about what environment you are in. the only thing i don't understand how to do is run debugging actions in my GUI (vscode). i may ask about that at some point but that will be a bit of a workflow change. in the meantime, let's keep this approach as an option and merge it! perhaps some won't want to use hatch to do their dev work and that is also OK.

PR #106, opened 2024-02-25
Fix: Add categories validator to model

### LINKED ISSUE #105: Fix: categories need a clean / validation step
We need a validator to clean up the categories attr in the pydantic model 

Current lists look like this: we want dashes between each word for easier parsing
```
    - data-retrieval
    - data-extraction
    - data-processing/munging
    - data-deposition
    - data-validation and-testing
    - workflow-automation
    - database-interoperability
```

will revisit that piece and potential pydantic validators to cleanup.

_Originally posted by @lwasser in https://github.com/pyOpenSci/pyosMeta/issues/93#issuecomment-1962236456_

lwasser:
closes #105 

Opening as a draft as it still needs tests but this adds a validation step to categories to ensure they are added to the final yaml file in a clean format with only dashes in between words for easier web parsing.

### DIFF
--- a/src/pyosmeta/parse_issues.py
+++ b/src/pyosmeta/parse_issues.py
@@ -265,6 +265,49 @@ class ReviewModel(BaseModel):
 
         return user
 
+    @field_validator(
+        "categories",
+        mode="before",
+    )
+    @classmethod
+    def clean_categories(cls, categories: list[str]) -> list[str]:
+        """Make sure each category in the list is a valid value.
+
+        Valid pyos software categories are:
+            citation-management-bibliometrics, data-retrieval,
+            data-extraction, data-processing-munging, data-deposition",
+            data-validation-testing, data-visualization-analysis,
+            workflow-automation-versioning, database-interoperability,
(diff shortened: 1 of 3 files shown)


src/pyosmeta/parse_issues.py
```diff
@@ -265,6 +265,49 @@ def clean_gh_url(cls, user: dict[str, str]) -> dict[str, str]:
...
+            List of categories to clean.
+
+        Returns
+        -------
+        list[str]
+            List of cleaned categories.
+        """
+
+        valid_categories = {
+            "data-processing": "data-processing-munging",
```
willingc, line 295: Small tip: If you alphabetize these categories, they will be easier to scan. Also in the docstring.

willingc (approved):

src/pyosmeta/parse_issues.py
lwasser: oh - good suggestion. i can fix that before we merge. it's a small fix and you're right more readable!

lwasser: fixed!

Issue #82, opened 2023-12-20
Migrate to VCS-based versioning using hatchling


lwasser:
Rather than static versioning, we should migrate pyosmeta to hatch_vcs for versioning and setup a GitHub action to trigger a release to pypi upon github release

* this also can be used as a practice run for our python packaging tutorial on this topic.

- [x] First, add vcs versioning to repo (see #85 )
- [x] #100

lwasser:
this works. i'll open a new issue to migrate to a more secure build .

PR #108, opened 2024-02-25
Fix: Update publish pypi workflow to use hatch

### LINKED ISSUE #100: Set up a github action using hatch to release to PyPI
Automate the release to PyPI using hatch.

To approach this issue, consult the hatch documentation and recommended GitHub action.

This task will finish and close #82.

lwasser:
this closes #100 #82

NOTE: i haven't set this up before with hatch but because hatch is using setuptools_scm i suspect it's the same workflow. let's see how this runs.

### DIFF
--- a/.github/workflows/publish-pypi.yml
+++ b/.github/workflows/publish-pypi.yml
@@ -46,10 +46,10 @@ jobs:
         with:
           # Haven't setup trusted workflow on test pypi yet
           password: ${{ secrets.TEST_PYPI_TOKEN }}
-          repository_url: https://test.pypi.org/legacy/
+          repository-url: https://test.pypi.org/legacy/
           # Allow existing releases on test PyPI without errors.
           # NOT TO BE USED in PyPI!
-          skip_existing: true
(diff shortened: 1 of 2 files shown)


.github/workflows/publish-pypi.yml
```diff
@@ -45,6 +44,7 @@ jobs:
         if: github.event_name == 'push'
         uses: pypa/gh-action-pypi-publish@release/v1
         with:
+          # Haven't setup trusted workflow on test pypi yet
```
lwasser, line 47: The weird part here is i'm not sure whether we want to use hatch publish vs just the pypi publish workflow which is easy and i've used before.

willingc: I think it is fine to keep pypa/gh-action-pypi-publish

willingc: Using hatch to build the wheel and sdist

lwasser:
merging. we can see if it goes to testpypi

PR #109, opened 2024-02-26
Fix grammar in devguide

willingc:
Missed this after adding your tutorial link suggestion.


lwasser (approved):
Thank you @willingc merging!

PR #110, opened 2024-02-26
Fix: test pypi trusted workflow setup, remove token

lwasser:
This is a tiny pr that should fix test pypi failing (as it did on @willingc last commit). if we merge this test pypi should recieve a new build of pyos meta (which is not published there yet).


lwasser:
i'm merging this to see if (test) pypi receives a new version

PR #112, opened 2024-02-26
Fix: add environment

lwasser:


.github/workflows/publish-pypi.yml
```diff
@@ -40,8 +42,8 @@ jobs:
           echo ""
           echo "Generated files:"
           ls -lh dist/
-
-      - name: Publish package on test PyPI on merge in main branch # optional
+
```
lwasser, line 45: here is what is odd about this.

i have a tag created for [0.2.1 ](https://github.com/pyOpenSci/pyosMeta/releases/tag/v0.2.1) which is the version that should get published to pypi. but we have some commits now on top of that version so i'm going to try to recreate it with all of the current commits.

willingc: I think the test version will get published to the test repo on every PR as written.

willingc: If each PR then it will be using the head of the branch.

PR #114, opened 2024-02-28
Fix: field alias for date_accepted field

lwasser:
Rather than fixing the entire thing i'm working towards cleaning up the parse_issues method. This is a small fix that 

1. adds a field alias to pydantic so there's not a random date field cleanup step in an unexpected place. 
2. adds a small test to ensure the both aliases defined in the model work 

i'm guessing there are better ways to setup tests like this but it's a start and i'm trying to keep this contained. this addresses #91 in (a very small) part.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 - Add: Code coverage setup (#97, @lwasser)
 - Fix: Add validation step to categories attr in pydantic model (#105, @lwasser)
 - Fix: update pypi publish to use hatch (#82, @lwasser)
+- Fix: move data accepted cleanup to pydantic field alias / validation step (@lwasser)
 
 ## [v0.15](https://github.com/pyOpenSci/pyosMeta/releases/tag/v0.15)
 
--- a/src/pyosmeta/cli/process_reviews.py
+++ b/src/pyosmeta/cli/process_reviews.py
@@ -29,7 +29,7 @@ def main():
         repo_name="software-submission",
         label_name="6/pyOS-approved 🚀🚀🚀",
     )
-
+    # date_accepted_(month/day/year)
     # Get all issues for approved packages - load as dict
     issues = process_review.return_response()
     # TODO: this method parse_issue_header is working but the parsing code is
--- a/src/pyosmeta/parse_issues.py
+++ b/src/pyosmeta/parse_issues.py
@@ -431,8 +431,9 @@ class ProcessIssues:
 
         Returns
         -------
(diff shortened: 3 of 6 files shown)


CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 - Add: Code coverage setup (#97, @lwasser)
 - Fix: Add validation step to categories attr in pydantic model (#105, @lwasser)
 - Fix: update pypi publish to use hatch (#82, @lwasser)
+- Fix: move data accepted cleanup to pydantic field alias / validation step (@lwasser)
```
lwasser, line 13:
```suggestion
- Fix: move data accepted cleanup to pydantic field alias / validation step (#91, @lwasser)
```

src/pyosmeta/cli/process_reviews.py
```diff
@@ -29,7 +29,7 @@ def main():
         repo_name="software-submission",
         label_name="6/pyOS-approved 🚀🚀🚀",
     )
-
+    # date_accepted_(month/day/year)
```
lwasser, line 32:
```suggestion
```

src/pyosmeta/parse_issues.py
```diff
@@ -572,9 +579,6 @@ def get_issue_meta(
         """
         issue_meta = {}
         for item in body_data[0:end_range]:
-            # Clean date accepted element
```
lwasser, line 575: this was just a thing i added in a really random place. it worked but it is so hard to find it if you're working on the code.

src/pyosmeta/parse_issues.py
```diff
@@ -161,7 +161,11 @@ class ReviewModel(BaseModel):
     reviewer_2: dict[str, str | None] = {}
     archive: Optional[str] = None
     version_accepted: Optional[str] = None
-    date_accepted: Optional[str] = None
+    date_accepted: Optional[str] = Field(
```
lwasser, line 164: here i added an alias to allow us to skip the step of cleaning up the field name

tests/unit/test_review_model.py
```diff
@@ -0,0 +1,36 @@
+from pyosmeta.parse_issues import ReviewModel
+
+# We could setup some example data using fixtures and a conf.py
+# Once we have a better view of the test suite.
+example = {
```
lwasser, line 5: we may end up creating fixtures for some of these things that we can reuse. but for now i just added a variable object to the test so we have a test for aliases

PR #115, opened 2024-02-29
Fix: make parse comment more readable

### LINKED ISSUE #91: Refactor needed for parse_issue_header method

lwasser:
closes #91 to make the method easier to follow and more readable. 
As it was there were multiple nested helpers making it super hard to follow / trouble shoot. i'm trying to surface each step and make each method name more user friendly with this draft pr.


lwasser:
ok i have added some very basic tests and some structure around the test suite. i don't want to bloat this pr even further so i think this is ready for a review / merge. ✨

PR #120, opened 2024-03-07
Fix: refactor and reorganize modules

lwasser:
This pr is an attempt to clean up and reorganize the various elements of this package. it does the following. 

1. it moves all of the pydantic models to a single models.py module. Then all of the pydantic things are only imported in that module. 
2. It moves the small cleaning functions that i created to a separate clean.py module. that way it's clear where they come from. 
3. it keeps the process_issue and process_contributors classes in separate modules by themselves. 

I have a question however. Right now #3 above - both modules need to call the github API. And as such i share the token authentation between the two classes. I wondered if it might make more sense to create a new github module that just handles github things? and i'd love some feedback on this. 

The goal here is to make the code easier to maintain and read.


willingc:
@lwasser Can you please provide a brief note for the PR on what changes are here ☀️

lwasser:
@willingc done!! i also have a question about organization.

willingc:
@lwasser Let's do this. Let me review this PR as is. We can then tackle simplifying the GitHub token use and naming in separate iterative PRs. Each PR can be small and tested with the goal of keeping the functionality working with minimal changes at once 👍🏼

src/pyosmeta/contributors.py
```diff
@@ -1,165 +1,10 @@
...
-    @classmethod
-    def clean_strings(cls, string: str) -> str:
-        """This is a cleaning step that will remove spurious
-        characters from string fields.
-
-        """
-        if isinstance(string, str):
-            string = re.sub(r"[\r\n]", "", string)
-        return string
+from typing import List, Optional, Tuple
```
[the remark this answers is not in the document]
willingc: Let's tackle this in a different PR.

lwasser:
@willingc that would be wonderful. i really appreciate your reviewing!!  🙌

lwasser:
i really see the value of smaller pr's. this one feels big big BUT it's mostly my moving things around and renaming. (and i removed some comments from my prior less create an issue savy self)

src/pyosmeta/clean.py
```diff
@@ -0,0 +1,104 @@
...
+    """Helper to strip unwanted chars from text"""
+
+    unwanted = ["(", ")", "@"]
+    for char in unwanted:
+        a_str = a_str.replace(char, "")
+
+    return a_str.strip()
+
+
+def parse_user_names(username: str) -> dict[str, str]:
```
willingc, line 42: Since this is not related to cleaning utilities, this may be better being moved to a file named `parse_utils` and rename this to `clean_utils`.

src/pyosmeta/clean.py
```diff
@@ -0,0 +1,104 @@
+"""Helpers that clean various markdown elements found in issue review text
```
willingc, line 1:
```suggestion
"""General cleaning utilities for markdown text.

We use these helpers to clean various markdown elements found in issue review text.
```

src/pyosmeta/models.py
```diff
@@ -0,0 +1,370 @@
+""""""
```
willingc, line 1:
```suggestion
"""Data models for Contributors, Reviews, and GitHub metadata.

This module also includes a convenience class for URL validation.
"""
```

willingc (approved):
Let's land the module docstring changes in this PR and leave renaming to future PRs. Thanks @lwasser. Nice work separating some of the code.

willingc (approved):
Thanks @lwasser. 🚢

PR #123, opened 2024-03-08
Fix: rename util modules

### LINKED ISSUE #121: Move `parse_user_names` to a new file called `parse_utils`
Since this is not related to cleaning utilities, `parse_user_names` may be better being moved to a file named `parse_utils` and rename `clean.py` to `clean_utils`.

_Originally posted by @willingc in https://github.com/pyOpenSci/pyosMeta/pull/120#discussion_r1517007443_

lwasser:
closes #121 
Here i moved parse_names to a new utils_parse.py module. i also renamed clean to utils_clean.py. i lead with utils to avoid confusing parse_utils with parse_issues and then the two utils modules sort nicely together in the package


willingc (approved):
Thanks @lwasser and great idea to prefix with `utils_`

PR #124, opened 2024-03-08
Add tests utils

### LINKED ISSUE #122: Add tests for utils module methods
there are a total of 4 methods in these modules. it will be quick to write unit tests for each to ensure they work as expected when cleaning and parsing our metadata.

lwasser:
closes #122 

this depends on pr #123 being merged first. it closes the loop on tests for all of the modules that we just moved around in 123.


willingc (approved):
Nice!

PR #126, opened 2024-03-08
Refactor cleaning utility input names to be consistent and update docstrings

willingc:
This PR improves the naming of function inputs for cleaning utilities by making the inputs and docstrings more generic.

Since all functions are used, this PR also removes the TODO in the module docstring.

This should help future contributors.

### DIFF
--- a/src/pyosmeta/utils_clean.py
+++ b/src/pyosmeta/utils_clean.py
@@ -7,20 +7,20 @@ import re
 from datetime import datetime
 
 
-def clean_date(a_date: str | None) -> datetime | str:
-    """Cleans up a datetime from github and returns a date string
+def clean_date(source_date: str | None) -> datetime | str:
+    """Cleans up a date string to a consistent datetime format.
 
-    In some cases the string is manually entered month-day-year and in
-    others it's a gh time stamp. finally sometimes it could be missing
-    or text. handle all of those cases with this validator.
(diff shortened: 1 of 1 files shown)


src/pyosmeta/utils_clean.py
```diff
@@ -1,28 +1,26 @@
...
-def clean_date(a_date: str | None) -> datetime | str:
-    """Cleans up a datetime from github and returns a date string
+def clean_date(source_date: str | None) -> datetime | str:
+    """Cleans up a date string to a consistent datetime format.
 
-    In some cases the string is manually entered month-day-year and in
-    others it's a gh time stamp. finally sometimes it could be missing
-    or text. handle all of those cases with this validator.
+    The source date string may have been manually entered as month-day-year format,
+    retrieved from GitHub as a timestamp, could be missing, or contain random text.
```
lwasser, line 14: Minor: right now as written it doesn't handle random text. Should we add a small fix issue to handle random text? there is a test for this now so the addition could then be added to the test and it could be an easy thing for an external contrib to tackle at a sprint too?

lwasser (approved):
This is great @willingc thank you. let's merge. i left a small question about opening an issue for the clean date function. we haven't had people adding random text to that field but i suppose it isn't a terrible idea to have a check for that too and return "missing".

PR #127, opened 2024-03-08
Split utilities test file into two to match the source files

willingc:
This small PR breaks up the existing test file for utilities into two test files following the refactor in #121 and #120. No tests were changed.

### DIFF
--- a/tests/unit/test_utils_clean.py
+++ b/tests/unit/test_utils_clean.py
@@ -1,13 +1,8 @@
-"""
-Tests for the clean and parse helper functions located in
-the utils_clean and utils_parse modules.
-
-"""
+"""Tests for the clean helper functions located in the utils_clean module."""
 
 import pytest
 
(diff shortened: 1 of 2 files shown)


lwasser (approved):
@willingc this is great. i debated about this last night as well. it feels weird to have a single function in a module and then a single test in a test module. are we leaning in to readibility over how much content is in a module here? we can merge this as well.

willingc:
@lwasser It's fine to have 1 function in a module or 1 test in a file. I tend to lean toward separation of concepts. If we had one file called `utils.py`, then I would create `test_utils.py` (which would include all utilities). Since we split the utils into two files which makes good sense `utils_clean.py` and `utils_parse.py`, I like to match the test files to the source file. It requires less thought on my part since I know my tests are almost always in a file `test_sourcefilename.py` when `sourcefilename.py` is the code file.

willingc:
Keep the great questions coming too.

lwasser:
Thank you @willingc !! That makes sense!!

Issue #113, opened 2024-02-27
Fix: publish to pypi build needs security enhancements


lwasser:
Following [this issue](https://github.com/pypa/gh-action-pypi-publish/issues/217) - we should make some changes to our release workflow to make it more robust. 

## Action items

- [x] Update action versions https://github.com/pyOpenSci/pyosMeta/actions/runs/8067048660 (specifically, the summary page, not inside the job, below the graph view), there are annotations. They contain 2 deprecation warnings that come from actions/checkout@v3, actions/setup-python@v2 — these action versions are old and may stop working at some point so it is a good idea to upgrade them sooner rather than later.
- [x]  break up our build and publish sections [as suggested here](https://github.com/pypa/gh-action-pypi-publish/issues/217#issuecomment-1965727093). This will create a more secure build. 

> an immediate recommendation — don't build your dists in the same job as publishing. Exposing the build machinery to the OIDC privilege is dangerous security-wise, as it may lead to privilege escalation in other systems through build dep poisoning. https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/ demonstrates good job separation

- [x] Remove test pypi from our build.  i've published to testpypi  in the past to make sure things still build and publish.  However test pypi doesn't maintainer current versions of packages so installing from there will never work - ie pydantic is super old on test pypi and unreliable (anyone could have uploaded it) 

Another route towards doing this is pre-releases. i just don't fully remember how to do that but have done it once for stravalib when we had a major API refactor / upgrade. 

- [ ] consider sigstore signing (maybe this is a new issue?)

lwasser:
I'm 100% on board with acknowledging them both in release notes @willingc as well as ofek as he's helped me a lot w hatch things. I'm so appreciative of all of the support in our community!!

webknjaz:
Since you're documenting follow-ups here, there's something I noticed but forgot to mention earlier. If you look at https://github.com/pyOpenSci/pyosMeta/actions/runs/8067048660 (specifically, the summary page, not inside the job, below the graph view), there are annotations. They contain 2 deprecation warnings that come from `actions/checkout@v3`, `actions/setup-python@v2` — these action versions are old and may stop working at some point so it is a good idea to upgrade them sooner rather than later.

[...]

PR #128, opened 2024-03-08
Update github actions to resolve nodejs 16 warnings

willingc:
Partially addresses #113 

This PR makes the versions used by checkout and python actions the same across all files. This should remove the warnings seen on some workflows about NodeJS 16 deprecation.


lwasser (approved):
thank you!!

PR #131, opened 2024-03-09
Fix: clean data accepted key

### LINKED ISSUE #129: BUG: date-accepted being removed from current build
If [you look here](https://github.com/pyOpenSci/pyopensci.github.io/pull/333/files#diff-2511ccad422b02cde04c39c34bf5e5c87952a50fe83f53ff56f8fa7ce43de6ffL689) you will notice that date-accepted is being REMOVED now from our current build. 

everything else looks good

if i had to guess, i know i made some changes to the validation / workflow of  of date_accepted. i thought it was all working but currently we don't have any tests that the output data in that model is as expected. SO... it looks like i broke something.

lwasser:
closes #129 

this fixes a bug in our workflow related to older issues with different formats for the date accepted key. it adds a helper clean function to the utils_clean module that looks for any key that begins with date_accepted and removes anything following it. this will simplify both our pydantic model ingest and also make it easy to add other scenarios if people modify that line.


willingc (approved):
Love this! Bugs being found and fixed in a place that makes sense to a reviewer.

willingc:
Looks like some sort of test issue.

lwasser:
yes! so essentially what i fixed here - there was a test that i didn't update. BUT i didn't see that it wasn't updated because it was not named correctly for pytest / vscode to find it  when i ran the tests. So i just fixed that all by 

1. fixing the test that should have been fixed
2. fixing the test suite setup (renaming the file to start with test_) so it ran when i ran tests locally. 
it's all green now - i shall merge given an approval. thank you so much! i'm happy with this fix as it's much easier to troubleshoot (and test) now.  ✨ 

thank you so much!

tests/integration/test_parse_issues.py
lwasser, line 1: this contained the failing test. it wasn't running for me locally and i didn't notice that it wasn't running. this fixes that issue and ensures i see broken tests before pushing to GH thinking everything is green!

tests/integration/test_review_model.py
```diff
@@ -18,7 +18,7 @@
     "archive": "[![DOI](https://zenodo.org/badge/DOI/10.5281/zenodo.8384174.svg)](https://doi.org/10.5281/zenodo.8384174)",
     "version_accepted": "5.1.1",
     "joss_doi": "[![DOI](https://joss.theoj.org/papers/10.21105/joss.01832/status.svg)](https://joss.theoj.org/papers/10.21105/joss.01832)",
-    "date_accepted_(month/day/year)": "01/18/2024",
+    "date_accepted": "01/18/2024",
```
lwasser, line 21: this was breaking the test because i fixed this key to ALWAYS be date_Accepted in this pr! so it needs to be fixed here too in the sample data.

lwasser:
all green ✅  merging! this one feels good as it feels like a good fix with a clean(er) fix implementation!

Issue #125, opened 2024-03-08
Refactor: Move all github things to new GitHub object


lwasser:
Currently the GitHub related API things are spread out between both the contributors and issue processing objects. One then ingests the other via instantiation / composition. 

This makes it hard to maintain github related functionality. And encourages duplication. 

It would be better to move that functionality out into a separate module called github_client. 
Then if we decide to move request to graphql (which might make sense in some calls) it would all happen in a single spot. 

So the first step of this is to simply 
1. move all github related methods into a new github_client module (with an object there)
2. Add a composition step to both issues and contributor objects where the github related methods are available. 
3. Swap out embedded functionality within each object to use the github object. 

This initial step should minimize if not remove any needed modifications fo the front end API. But will make the back end easier to maintain / update / debug.

PR #132, opened 2024-03-09
Fix: Move all of the github things to a new module for easier maintenance

lwasser:
this is the first part of addressing #125 

it moves all of the gh methods from parse issues into a new module. 
tests are still needed so opening as a draft.

### DIFF
--- a/src/pyosmeta/contributors.py
+++ b/src/pyosmeta/contributors.py
@@ -1,7 +1,9 @@
 import json
+import os
 
 import requests
 from dataclasses import dataclass
+from dotenv import load_dotenv
 from typing import List, Optional, Tuple
 
 
@@ -50,17 +52,17 @@ class ProcessContributors:
             ],
         }
 
-    # def get_token(self) -> str:
-    #     """Fetches the GitHub API key from the users environment. If running
-    #     local from an .env file.
+    def get_token(self) -> str:
+        """Fetches the GitHub API key from the users environment. If running
+        local from an .env file.
 
-    #     Returns
-    #     -------
-    #     str
-    #         The provided API key in the .env file.
-    #     """
-    #     load_dotenv()
-    #     return os.environ["GITHUB_TOKEN"]
+        Returns
+        -------
+        str
+            The provided API key in the .env file.
+        """
+        load_dotenv()
+        return os.environ["GITHUB_TOKEN"]
 
     def check_contrib_type(self, json_file: str):
         """
--- a/src/pyosmeta/contributors.py
+++ b/src/pyosmeta/contributors.py
@@ -1,9 +1,7 @@
 import json
-import os
 
 import requests
 from dataclasses import dataclass
-from dotenv import load_dotenv
 from typing import List, Optional, Tuple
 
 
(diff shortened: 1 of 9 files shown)


src/pyosmeta/contributors.py
```diff
@@ -1,9 +1,7 @@
 import json
```
lwasser, line 1: i am going to remove this file from this PR as we will handle that in another pr. this will break our contributor workflow as it is now

src/pyosmeta/github_api.py
```diff
@@ -0,0 +1,216 @@
...
+import os
+
+import requests
+from dataclasses import dataclass
+from dotenv import load_dotenv
+from typing import Any
+
+
+@dataclass
+class GitHubAPI:
```
willingc, line 22: I wonder if this makes more sense to name GitHubMetadata? Anything dealing with metadata from GitHub could then go into this file. Maybe renaming the file to `github_metadata.py`.

PS @lwasser I love your approach and how you are thinking about how the next developer will react to the code.

src/pyosmeta/parse_issues.py
```diff
@@ -392,80 +350,60 @@ def get_gh_metrics(
...
-        else:
-            return len(response.json())
+        stats_dict = {}
+        # Returns the raw top-level github API response for the repo
+        gh_repo_response = self.github_api.get_repo_meta(url)
 
-    def get_last_commit(self, repo: str) -> str:
-        """Returns the last commit to the repository.
+        # Retrieve the metrics that we want to track
+        for astat in self.gh_stats:
```
willingc, line 401: Perhaps `stat` instead of `astat`

src/pyosmeta/parse_issues.py
```diff
@@ -392,80 +350,60 @@ def get_gh_metrics(
...
+            stats_dict[astat] = gh_repo_response[astat]
```
willingc, line 402: You might consider not adding type into the variable name as it can be confusing and change over time. `stats[stat]` is more readable since I look at `stats_dict` and ask myself why if this is a dict is it formed like a list?

willingc: It's fine to leave as is for this PR. But may be something to consider iterating on in future PRs. 🌻

lwasser: hi @willingc 👋 i will make these changes!  

i'm getting hung up on the testing. Trying to create mock environments is not going as planned. Should i hold off on tests for a new pr so this one could be merged once we're happy with naming and such? i find API / environment related tests to be tricky to setup.

src/pyosmeta/github_api.py
lwasser: Right now i'm only moving methods into this module that interface with the github API. and then all of the actual metadata parsing is happening in the issues and contributors modules. so i'm a bit hesitant on naming it metadata here as well. this module is for authentication and get requests to the API only (as i've moved things anyway now). this will be the hardest module to write tests for as well. 😆

lwasser:
weird. the test that won't run on my computer runs here. i don't understand temp directories and pytest. it should run in both places.

lwasser:
ok i am going to merge this pr. tests are passing, the red x is code cover is unhappy but we are moving forward in test coverage even if code code is unhappy! the next PR will be to migrate all of the github API calls from process_contribs to the new ghapi module!

PR #138, opened 2024-03-18
Fix: refactor github methods out of contributor mod

### LINKED ISSUE #125: Refactor: Move all github things to new GitHub object

lwasser:
This should close #125 it's the second half of migrating all of our github methods into the github_api module. 

It also adds some docstrings and clarification in some of the methods. 

i am going to add a few tests as well once we know everything is passing as is.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,8 +9,9 @@
 - Fix: Rename and organize `clean.py` module into `utils_parse` and `utils_clean` (@lwasser, @willingc, #121)
 - Fix: Add tests for all utils functions (@lwasser, #122)
 - Fix: Bug where date_accepted is removed (@lwasser, #129)
-- Fix: Refactor all GitHub related methods move to gh_client module (@lwasser, #125)
+- Fix: Refactor all issue related GitHub methods to gh_client module (@lwasser, #125)
 - Add: support for partners and emeritus_editor in contributor model (@lwasser, #133)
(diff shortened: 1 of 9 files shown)


lwasser:
@willingc this PR got a lot bigger than i expected. I tried to add some tests which ended up including some fixtures which ended up leading to my finding that docstrings and such were missing / unclear. i suspect this is not one of my better / more clean pr's so i would appreciate any high level feedback to make it more digestable. or for just future pr's being more digestable.

willingc (approved):
Looks good to me. Thanks @lwasser 

P.S. The PR length was fine. It was easy to review since it was all related to the same topic/behavior. Nice work!

lwasser:
ok thank you so much @willingc !!! i'll merge this now! i'm glad to hear you felt it was somewhat reasonable in terms of volume of changes. it was pretty hard to scope this one!

PR #153, opened 2024-05-07
Fix: bug in github username for reviewers)

lwasser:
This is a small bug fix. this action hasn't been running in a few weeks because of this.

### DIFF
--- a/src/pyosmeta/cli/update_review_teams.py
+++ b/src/pyosmeta/cli/update_review_teams.py
@@ -35,6 +35,11 @@ from pyosmeta.models import PersonModel
 def get_clean_user(username: str) -> str:
     """A small helper that removes whitespace and ensures username is
     lower case"""
+    # If the github username has spaces there is an error which might be
+    # that someone added a name after the username. Only return the github
+    # username
+    if len(username.split()) > 1:
+        username = username.split()[0]
     return username.lower().strip()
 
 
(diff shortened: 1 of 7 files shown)


src/pyosmeta/cli/update_review_teams.py
```diff
@@ -30,12 +30,7 @@
 from pyosmeta.file_io import clean_export_yml, load_pickle
 from pyosmeta.github_api import GitHubAPI
 from pyosmeta.models import PersonModel
-
-
-def get_clean_user(username: str) -> str:
```
lwasser, line 35: moved this to the utils module as it makes more sense there.

src/pyosmeta/cli/update_review_teams.py
```diff
@@ -61,7 +56,9 @@ def main():
 
                     if gh_user not in contribs.keys():
                         print("Found a new contributor!", gh_user)
-                        new_contrib = process_contribs.get_user_info(gh_user)
+                        new_contrib = process_contribs.return_user_info(
```
lwasser, line 59: This was the cause of the actual first bug in CI

src/pyosmeta/utils_clean.py
```diff
@@ -9,6 +9,39 @@
...
+    Examples
+    --------
+    >>> get_clean_user("@githubusername")
+    'githubusername'
+
+    >>> get_clean_user("@githubusername name after text")
+    'githubusername'
+    """
+
+    if len(username.split()) > 1:
```
lwasser, line 40: Once the above bug was fixed another bug popped up related to a github username entered with a name after (and only a space).

PR #156, opened 2024-05-25
Sign pypi releases using sigstore

### LINKED ISSUE #146: Sign pypi releases using sigstore
In [this issue](https://github.com/pyOpenSci/pyosMeta/issues/113#issuecomment-1967958803) @webknjaz suggested looking into sigstore as a way to sign releases and add additional security to our builds. myself and @willingc have  implemented everything else in that issue short of sigstore. i'll save this for a future release / effort! Notes on it below:

> Finally, there's a way to sign releases with Sigstore that's showcased @ https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/#signing-the-distribution-packages (might need adapting to take your trigger into account). This is optional but with the support for uploading GPG signatures removed from the PyPI, it's probably going to be the future replacement so I decided to include a configuration example that also relies on OIDC but doesn't need much configuration.

blink1073:
Fixes #146 

Uses the example given at [packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/#signing-the-distribution-packages](https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/#signing-the-distribution-packages).

I adapted it for the use case of this repository.  The major difference in workflow is that this repository starts the release process by creating a GitHub Release, while the referenced example creates the GitHub Release as part of the workflow.

The new job runs after the `publish` step has completed, to ensure that the files have been correctly uploaded to PyPI.  It is run in a dedicated `signature` environment, so that it can have its own set of environment permissions and secrets, if desired.  The job is only run against the source repository, not on forks.

The job downloads the distribution files that were created in the `build` job, and then uses the `sigstore/gh-action-sigstore-python` action to create signature files for the distribution files.  This action requires on a GitHub-provided OIDC token, which is why the job has `id-token: write` permissions.

Finally, the job uploads the distribution and signature files to GitHub Release that triggered the workflow.    This is only done if the workflow was trigged by a release event.   This action requires `contents:write` permission to be able to upload the files to the GitHub release.

--- 

I also added a new pre-commit hook that validates GitHub workflows against the published JSON schema.

### DIFF
--- a/.github/workflows/publish-pypi.yml
+++ b/.github/workflows/publish-pypi.yml
@@ -70,3 +70,38 @@ jobs:
         # Only publish to real PyPI on release
         if: github.event_name == 'release'
         uses: pypa/gh-action-pypi-publish@release/v1
+  sign-files:
+    name: >-
+      Sign the Python 🐍 distribution 📦 with Sigstore
+      and upload them to GitHub Release
+    if: github.repository_owner == 'pyopensci'
+    needs:
+    - publish
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write  # this permission is mandatory for modifying GitHub Releases
+      id-token: write  # this permission is mandatory for sigstore
+    steps:
+    - name: Download all the dists
+      uses: actions/download-artifact@v3
+      with:
+        name: python-package-distributions
+        path: dist/
+    - name: Sign the dists with Sigstore
+      uses: sigstore/gh-action-sigstore-python@v2.1.1
+      with:
+        inputs: >-
+          ./dist/*.tar.gz
+          ./dist/*.whl
+    - name: Upload artifact signatures to GitHub Release
+      # Only upload on release
+      if: github.event_name == 'release'
+      env:
+        GITHUB_TOKEN: ${{ github.token }}
+      # Upload to GitHub Release using the `gh` CLI.
+      # `dist/` contains the built packages, and the
+      # sigstore-produced signatures and certificates.
+      run: >-
+        gh release upload
+        '${{ github.ref_name }}' dist/**
+        --repo '${{ github.repository }}'
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -56,3 +56,9 @@ repos:
       - id: isort
         files: \.py$
         args: ["--profile", "black", --line-length=79]
+
+  # Ensure GitHub workflows match the expected schema.
+  - repo: https://github.com/sirosen/check-jsonschema
+    rev: 0.28.3
+    hooks:
+      - id: check-github-workflows


blink1073:
I'm happy to change it to use the same `pypi` environment as the publish job if desired, otherwise someone with admin access will need to create the new environment.  :smile:

lwasser:
@blink1073 thank you so much for this pr!! i'll will have a look at it this week! i am a bit behind :)

lwasser:
@all-contributors please add @blink1073 for code, review

.pre-commit-config.yaml
```diff
@@ -56,3 +56,9 @@ repos:
       - id: isort
         files: \.py$
         args: ["--profile", "black", --line-length=79]
+
+  # Ensure GitHub workflows match the expected schema.
+  - repo: https://github.com/sirosen/check-jsonschema
```
lwasser, line 61: Wow - this is really cool. i've definitely pushed files with yaml syntax errors before.

lwasser (approved):
thank you @blink1073 i'm going to merge this and test out a release!! it's so helpful to have CI help in our org!!

lwasser:
> I'm happy to change it to use the same `pypi` environment as the publish job if desired, otherwise someone with admin access will need to create the new environment. 😄

@blink1073 can you please help me understand this? I could create an environment OR give you access! 

the environment piece within a CI action was new to me here - essentially @webknjaz pointed out that we weren't using best security practices in publishing to pypi! But i don't fully understand where the environment would live outside of telling the action to create an environment and then specifying that environment on pypi. many thanks.

webknjaz:
@lwasser GitHub Environments is a way to represent deployment targets within GH UI/ecosystem. It's tied to the Deployments API under the hood. When you use a non-existing GH Environment in GHA, it's autocreated. It can also be created manually. Once it exists, you can navigate to the repository settings and configure protections like required reviews, a cooldown timer etc. When set, this is able to pause a GHA workflow right before starting a job marked with an env — a manual button click on the UI would be needed to let it start, giving you more control over letting immutable changes happen on PyPI, for example. If you set the `url:` key, it also displays a nice link in the workflow graph, that URL can even interpolate variables so you could construct a precise versioned string. The deployments also have a separate web page in the repo.
When using trusted publishing, GitHub will present the GH env name as a part of signed piece of metadata that's being sent to PyPI. And so PyPI is able to know and trust that an env with a specific name was used when publishing is requested. You can reconfigure trust on the PyPI side to only allow uploads from jobs where a specific GH env name is set. That's what I suggest doing.
This is a part of my guide at https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/ by default.

.pre-commit-config.yaml
webknjaz: There's a few more related schemas there + the ability to host a custom one locally. Though, sometimes those schemas have bugs too.

.github/workflows/publish-pypi.yml
```diff
@@ -70,3 +70,38 @@ jobs:
...
+        path: dist/
+    - name: Sign the dists with Sigstore
+      uses: sigstore/gh-action-sigstore-python@v2.1.1
+      with:
+        inputs: >-
+          ./dist/*.tar.gz
+          ./dist/*.whl
+    - name: Upload artifact signatures to GitHub Release
+      # Only upload on release
+      if: github.event_name == 'release'
```
webknjaz, line 98: It probably makes sense to put this into the job conditional. Otherwise, this may end up signing undesired files, which may be used in supply chain attacks.

webknjaz (commented):
I think this change looks good in general. Except for that step condition. Hopefully, the attestations will start being uploaded as part of the trusted publishing mode later this year, once William finishes his PRs...

webknjaz:
@lwasser I didn't look through the PR before writing the first comment. Now that I've done so, I think your use of envs is good. The new env might be something to configure optionally, but it'll work even without that.

lwasser:
ok so @webknjaz thank you! i didn't fully understand that the environment was autocreated when you run the action. 
Screenshot 2024-06-05 at 3 13 17 PM

But i see them now in the settings. so that is why @blink1073 was talking about admin rights. 
So the new build environment is also there in GitHub. 

It sounds like you are saying that it is ok to have both the pypi environment (which is used in PyPI.org to allow the trusted publisher workflow to run) and the build environment that steven added to support sigstore publishing. 

Thank you. there is a lot to understand here!!

.github/workflows/publish-pypi.yml
blink1073: Done

blink1073:
Thanks for the review both of you!  I updated the job to use the `pypi` environment.

webknjaz:
> Thanks for the review both of you! I updated the job to use the `pypi` environment.

I don't think that should be the same, since this is putting artifacts into a deployment target that is not PyPI. If you want to declare an environment, I'd call it `gh-releases` or `sigstore`, maybe. However, it's probably not necessary to mark the job with an environment. If `pypi` is configured to pause before starting jobs, it will pause before each, but the signing job only runs after publishing, so pausing it there will already prevent it from running until publishing is approved. And if publishing happened, there's no reason to prevent signing that follows it.

webknjaz:
> It sounds like you are saying that it is ok to have both the pypi environment (which is used in PyPI.org to allow the trusted publisher workflow to run) and the build environment that steven added to support sigstore publishing.

It would only matter if you were to do something that is configured within an environment. Like, if you were to have an environment-scoped secret (as opposed to a repo-global one) for some reason, I'd recommend using it. But as things stand, it's probably overengineering, and I'd just remove `environment:` from this job.

blink1073:
> But as things stand, it's probably overengineering, and I'd just remove environment: from this job.

Yeah that's fair, done.

Issue #168, opened 2024-07-01
Broken build - yet again


lwasser:
This package is used to update our contributor info on the website. Right now the build is broken - again.

Please [see here](https://github.com/pyOpenSci/pyopensci.github.io/actions/workflows/update-contribs-reviews.yml).

@blink1073 if you happen to be looking for another CI issue to fix this one might include some python. i'm also happy to give you merge permissions if that is of interest!! this build seems to fail periodically. most often it's because someone adds something to our review template that i hadn't expected when parsing the issues! But it's important as it keeps our website [packages](https://www.pyopensci.org/python-packages.html) listing & [contributor](https://www.pyopensci.org/our-community/index.html#pyopensci-community-contributors) list up to date.

lwasser:
a new release might fix this issue.

blink1073:
Yes, it looks like it needs #166

blink1073:
I'm happy to accept write permissions, and also look for a way to test this workflow as part of PRs.

blink1073:
PR: https://github.com/pyOpenSci/pyosMeta/pull/170

lwasser:
sweet. i just made you a maintainer of this repo and added you to our org. i can add you to others as needed too! please just let me know what permissions you need where!

blink1073:
will do, thanks!

lwasser:
And now that ci build is helping me troubleshoot here before the website build breaks. thank you @blink1073 there is one more bug to resolve but i'll do that in a separate issue and will close this one!!

PR #170, opened 2024-07-01
Test cli on pull requests

blink1073:
Addresses #168

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -5,6 +5,9 @@ on:
   push:
     branches:
       - main
+  schedule:
+    # Runs on Mon and Thu of each week at 00:00 UTC (see https://crontab.guru)
+    - cron: "0 0 * * 1,4"
 
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


lwasser:
thank you!!

lwasser (approved):

.github/workflows/run-tests.yml
```diff
@@ -21,6 +21,17 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: hatch run test:run-report
+      - name: Run CLIs against website
```
lwasser, line 24: could we run this as a cron job as well? i ask because often what happens is that there is an issue in how a user fills out an issue. so they do something wierd (text wise) and it breaks the build. i was thikning if we ran this every week or twice a week that would catch those issues before we go to update contributors (i think?) i'm not sure if that thought process makes sense or not...

 but in this case i definitely broke something via a pr and this will fix that instance!

blink1073: done!

lwasser:
thank you @blink1073 !! i'm going to merge this and then create a new release. in the future you can just merge (if you want!) !!

Issue #186, opened 2024-07-11
bug: Problem with how partners field is parsed + tests


lwasser:
Right now, the partners field for astropy specifically, looks like this as an output:

https://github.com/pyOpenSci/pyopensci.github.io/blob/main/_data/packages.yml#L81

```yaml
  partners:
    - 'astropy: link-coming soon-to standards'
``` 

this is breaking the build on the astropy side of things - [see here](https://www.astropy.org/affiliated/#affiliated-package-list). 

the partners field should always serialize to single word partner names like this:

```yaml
  partners:
    - 'astropy: link-coming soon-to standards'
``` 

note: We should add some tests to ensure that the model captures / serializes fields correctly. let's start with a test for the partners field when this is fixed.

cc: @pllim

pllim:
Thanks! Does this mean my patch over at https://github.com/astropy/astropy.github.com/pull/605 is irrelevant or still applies?

lwasser:
@pllim your patch accounts for this bug. but once we fix this bug your patch won't be necessary if that makes sense. so if you are ok with it potentially taking a bit of time to fix this bug depending on who has time -- (i'm at scipy) - then you dont need that patch! 

we should be able to fix this quickly once we have time. we just need to make the partners field a Literal as @sneakers-the-rat mentioned and a validator that just uses "starts with" i think the check / clean / ingest the correct partner value

pllim:
Well, the website isn't showing anything is kinda shocking, so I think I will just merge what I have downstream so things are fixed for now. If the patch is compatible with your future fix, then it does not hurt anyway, right? Thanks for the advice!

lwasser:
nope - it will NOT hurt to merge that! i'm sorry about this complication. we are working on tests for this build so we know about these changes in the future BEFORE they are ingested into our website build!!

PR #187, opened 2024-07-12
fix parsing of partnerships

sneakers-the-rat:
Fix: https://github.com/pyOpenSci/pyosMeta/issues/186

sorry this is coming so late today, i have been helping everyone else will stuff all day long and haven't even gotten to any of my own work yet lmao, but since i caused this bug i wanted to fix ASAP.

so here's fix as discussed 
- **model** - used  string enum (python 3.10 doesn't have `enum.StrEnum`) instead of a `Literal` because in the future it would be nice to be able to add annotations like the text that comes after the key in the template currently so that we can generate the review template directly from the model. It does the same thing - the pydantic model will validate and throw an error there if there's any value that isn't in that enum. 
- **get_categories** - since the same method `get_categories` is used for parsing a regular checklist and a checklist that is effectively a key: value pair, i added a `keyed` param that extracts the key. 
- **test, regex** - to do that, i made a regex pattern and added tests for that so at least i have declared the expected behavior of that pattern lol
- **test, parsing** - also added a fake review based on a real one to test actual parsing of an actual issue

so by making the model more strict we now can make the claim that "this key will always be this value or absent" and "if the string is within the tested boundaries of the regex pattern it will be present." We need to do that for the rest of the model now too - make it reflect the contract we want to make with anyone ingesting data from our packages.yaml file: eg. currently `package_name` is optional and defaults to a blank string, but we should actually never have a review with no package name. and so on.  but that's enough for this v simple lil bug

### DIFF
--- a/src/pyosmeta/models/base.py
+++ b/src/pyosmeta/models/base.py
@@ -20,6 +20,7 @@ from pydantic import (
 
 from pyosmeta.utils_clean import clean_date, clean_markdown
 
+
 class Partnerships(str, Enum):
     astropy = 'astropy'
     pangeo = 'pangeo'
--- a/src/pyosmeta/models/base.py
+++ b/src/pyosmeta/models/base.py
@@ -22,8 +22,9 @@ from pyosmeta.utils_clean import clean_date, clean_markdown
 
 
 class Partnerships(str, Enum):
-    astropy = 'astropy'
-    pangeo = 'pangeo'
+    astropy = "astropy"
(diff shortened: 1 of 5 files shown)


lwasser:
@sneakers-the-rat thank you for this!! it works perfectly. i'm going to merge now.

lwasser:
you are also correct there are a few please where we could impose a `Literal` // set of accepted values. And also not accept a missing field such as package name and repo url (authors / maintainers too).

Issue #188, opened 2024-07-12
Add: Review model should include labels (OR is there a better way to access labels)? 


lwasser:
Current we have a peer-review-metrics repo that supports gathering information on our current and past review status. The goal is to turn this into a dashboard for editors and the EiC to understand the current state of reviews while also collecting review metrics that I / we can use for reporting. 

The current refactor is great but it broke a few things over in peer-review metrics (built with myst). 

see this [notebook as an example](https://github.com/pyOpenSci/peer-review-metrics/blob/main/current-reviews-data.ipynb). One of the things that we want to access with our review issues is the labels. The labels tell us the status of a review, and we need that information to create our dashboard of current and previous reviews. 

we also want to collect metadata on the issue such as date opened, date closed, etc.

I wonder if we create a separate object or add a subclass to ReviewModel that stores Issue metadata including 

* labels
* date_opened
* date_closed  

We may want to add to this in the future as well. 
@sneakers-the-rat what do you think makes sense structurally here for this?

Would something like 
```python
ReviewModel:
    issue_meta: IssueMeta
    
IssueMeta:
     labels: list[str]
     date_opened: datetime 
     date_closed: datetime
 ```
Make sense ? 

i wasn't thinking about our little dashboard that i've been playing with when we merged! but i think your structure is still much much cleaner!
      

    
make sense as a subclass that captures metadata like labels,

sneakers-the-rat:
sure, could definitely be done. we make the model how we want it to work! ideally we do any big changes like this before anyone is actually depending on the fine structure of it (idk how many people actually are, but if there are many then we would want some deprecation process). 

we already have [this](https://github.com/pyOpenSci/pyosMeta/blob/e0fcb08800fd4456fb1ce3058d393286fddf9cc0/src/pyosmeta/parse_issues.py#L157) one liner for directly copying fields from the Issue, so eg. we already have `created_at` and `closed_at` which come from the `Issue` model, i'll PR grabbing labels now. 

semantically i guess all of the contents of the `ReviewModel` come from an issue so are technically `IssueMeta`, but if it would be helpful to group the fields that we are directly copying from the issue api response then by all means!

lwasser:
Thank you so much @sneakers-the-rat ! If it helps - right now this package supports:

1. generating that yaml file that populates our website and 
2. That peer review metric "dashboard" i've been working on just as a side thing.

So those two builds are the only ones that our changes will impact given this is a specific-to-pyOpenSci package.  

[...]

PR #189, opened 2024-07-12
get labels from issue metadata

sneakers-the-rat:
Fix: https://github.com/pyOpenSci/pyosMeta/issues/188

sorta self explanatory, we can keep the whole `Issue` object if we want, i just figure we sorta want to keep the review `yaml` tidy-ish and wasn't sure if we would want to use any of those other fields. if we want it it would just be a matter of removing the validator and changing the field type.


lwasser:
This also works great! thank you again. I see the label names now in the issue object. 
i'm still wrapping my head around using Field and mutable vs immutable objects!!  merging.

PR #193, opened 2024-07-13
Fix: fix label parsing

### LINKED ISSUE #192: [bug / feature] : challenges pulling all review issues from our pyos meta package
Right now i'm working on our dashboard for peer review metrics. 

if we run this code to collect all issues:

```python
from pyosmeta import ProcessIssues
from pyosmeta.github_api import GitHubAPI

# Get all issues from GitHub software-submission repo, Return df with labels, title, date_opened and closed and total time open in days
github_api = GitHubAPI(
    org="pyopensci",
    repo="software-submission",
)

process_review = ProcessIssues(github_api)
issues = process_review.get_issues()
accepted_reviews, errors = process_review.parse_issues(issues)
```

It errors at a string/split step (i haven't had time to fully debug yet). 
There are going to often be times when we have issues with odd characters in them. In this case because it's parsing ALL issues, there are likely some issues that are not related to peer review that it's getting stuck on OR it's getting stuck on presubmissions, etc. 

There are a few options here

1. We always use a label to grab issues or we provide a list of labels (reasonable). 
2. We allow the parses to fail gracefully on an issue, and report back the issue numbers of what it skips. 

because i haven't looked into this more, i'm not sure what the best approach is at this point. But i wanted to log the issue as i work on peer-review-metrics!

lwasser:
closes #192 

Right now i was trying to pull all issues which meant it was grabbing issues that were NOT actually review issues. I decided it was better to be explicit. This pr thus fixes two things

1. it allows us to parse through issues and look for issues that contain a specific label. By default the rest endpoint returns issues with ALL labels included. But we wanted the select an issue based upon a specific label occuring within a set of labels. This implements that by filtering issues. GraphQL i think has an OR option? but i just am trying to get the peer-review-metrics running again with this pr. 
2. This pr also adds an alias for repo_link to support the presubmission template

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ## [Unreleased]
 
+* Fix: allow for issues with multiple labels & fix presubmission ingest (@lwasser)
+
 ## [v0.3.2] - 2024-07-04
 
 ### Fixes
--- a/src/pyosmeta/github_api.py
+++ b/src/pyosmeta/github_api.py
@@ -73,12 +73,34 @@ class GitHubAPI:
             )
 
     @property
-    def api_endpoint(self):
-        labels_query = ",".join(self.labels) if self.labels else ""
-        url = (
-            f"https://api.github.com/repos/{self.org}/{self.repo}/"
-            f"issues?labels={labels_query}&state=all&per_page=100"
-        )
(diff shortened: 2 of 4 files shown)


lwasser:
I am merging and creating a release quickly to see if i can fix the build before our sprints tomorrow!

PR #200, opened 2024-08-01
fix: advisory_council should default to false

lwasser:
This is a small fix to ensure that emeritus advisory and advisory roles:

1. default to false
2. always have a bool value which should default to false
3. should not be changed by the code if the value in the data is true (i think this was just a tiny bug that i found).

### DIFF
--- a/src/pyosmeta/models/base.py
+++ b/src/pyosmeta/models/base.py
@@ -104,12 +104,12 @@ class PersonModel(BaseModel, UrlValidatorMixin):
     editorial_board: Optional[bool] = Field(
         None, validation_alias=AliasChoices("editorial-board")
     )
-    emeritus_editor: Optional[bool] = Field(
-        None, validation_alias=AliasChoices("emeritus_editor")
+    emeritus_editor: bool = Field(
+        False, validation_alias=AliasChoices("emeritus_editor")
     )
-    advisory: Optional[bool] = False
-    emeritus_advisory: Optional[bool] = Field(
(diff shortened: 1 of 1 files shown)


src/pyosmeta/models/base.py
```diff
@@ -130,6 +130,19 @@ class PersonModel(BaseModel, UrlValidatorMixin):
     location: Optional[str] = None
     email: Optional[str] = None
 
+    @field_validator(
+        "emeritus_advisory",
+        "advisory",
+        "emeritus_editor",
+        "editorial_board",
+        mode="before",
+        always=True,
```
willingc, line 139: @lwasser It looks like this field is causing an issue with CI. What is the purpose of `always`?

lwasser: oh shoot @willingc i merged this too quickly. in the PR the tests had passed  (well i thought they had maybe they hadn't!) I suspect always=true was made up by LLMS. i often use chatgpt to proof my code / format etc. and it suggested that addition and i used it without researching. this is my mistake. i should know better than to listen to a llm!

lwasser: i looked more closely. one had failed. i missed it. - https://github.com/pyOpenSci/pyosMeta/actions/runs/10201975001/job/28225084163#step:5:28 i'll submit a fix and won't be so rushed this time.

willingc: Cool! No problem.

PR #203, opened 2024-08-01
Fix: model bug that i introduced

lwasser:
In this pr I: 

1. fix the bug that i introduced with the field validator
2. update the changelog to capture the last few commits / prs


lwasser:
ok i did test this locally. AND everything looks good. It also is GREEN passing all tests. I am going to merge this. we should be ready for a release now.

PR #211, opened 2024-08-14
feat: support for issues ++ prs ✨  in pyosmeta

lwasser:
this is a draft issue that adds support for both issues and pr's into pyos meta. this will allow us to more easily process metrics that we need for reporting.


lwasser:
I am going to merge this and create a tiny patch release to support this feature for our metric generation that I need for the next proposal // annual report.

Issue #227, opened 2024-11-19
pyosmeta doing odd things with the gitub username / name


lwasser:
These three issues are NOT related so i'm breaking them out.  

- [x] DONE - 1. The user's name should always remain what is in our yml file and not get updated once it is populated. If it's fully missing then we can update it using github. It also should never be Name 😆 
[here](https://github.com/pyOpenSci/pyopensci.github.io/pull/527/files#diff-2511ccad422b02cde04c39c34bf5e5c87952a50fe83f53ff56f8fa7ce43de6ffR49) we have an example that looks like this
Update: this was a glitch. The word "Name" was actually in the review issue - a typo / so i fixed that
 
[see here](https://github.com/pyOpenSci/pyopensci.github.io/pull/520/files#diff-2511ccad422b02cde04c39c34bf5e5c87952a50fe83f53ff56f8fa7ce43de6ffL619) for an example. 
<img width="1142" alt="Screenshot 2024-11-20 at 6 46 25 PM" src="https://github.com/user-attachments/assets/dfd184e5-7fc3-4011-b84e-e6deb11b8f48">
FIXED 🚀 

- [x]  [This](https://github.com/pyOpenSci/pyopensci.github.io/pull/527/files#diff-2511ccad422b02cde04c39c34bf5e5c87952a50fe83f53ff56f8fa7ce43de6ffR64) is an example of pyosmeta completely erasing a name. This one baffles me as Chiara has been with us for some time, and this hasn't happened before in our bot pro.  This issue is also FIXED! it was because of #234, where the eic field wasn't being parsed, so it cleared out the name because the name wasn't being parsed properly. 

Note =: the pr may take a bit of time to merge as it would be nice to have tests. 
 🚀 

*******

- [x]  🆘  Username should never contain the GitHub handle 🆘  this is not yet fixed!! 

[this](https://github.com/pyOpenSci/pyopensci.github.io/pull/527/files#diff-2511ccad422b02cde04c39c34bf5e5c87952a50fe83f53ff56f8fa7ce43de6ffR657)
is an example of a user's name, including the GitHub username too, in our data. that is also odd and shouldn't happen.

lwasser:
i'm going to open a  sub issue

lwasser:
this can now be closed as resolved. it's due to a bug in the template that i forgot about

PR #231, opened 2024-11-21
Fix names

lwasser:
This issue documented how a user's information is updated in our workflow. it's all inline documentation.   

it begins to tackle #227 but it doesn't resolve it. it just resolves that question around - how is the name populated and why is the name sometimes being removed?

### DIFF
--- a/src/pyosmeta/cli/update_review_teams.py
+++ b/src/pyosmeta/cli/update_review_teams.py
@@ -1,10 +1,11 @@
 """
-This script parses through our reviews and contributors and:
+This script parses through our packages.yml and contributors.yml files
+and:
 
 1. Updates reviewer, editor and maintainer data in the contributor.yml file to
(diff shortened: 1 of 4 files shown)


src/pyosmeta/cli/update_review_teams.py
```diff
@@ -1,10 +1,11 @@
 """
-This script parses through our reviews and contributors and:
+This script parses through our packages.yml and contributors.yml files
+and:
 
 1. Updates reviewer, editor and maintainer data in the contributor.yml file to
 ensure all packages they supported are listed there.
 1b: And that they have a listing as peer-review under contributor type
-2. Updates the packages metadata with the participants names if it's missing
+2. TODO: not working as expected? Updates the `packages.yml` file with review contributor's names if it's missing
```
lwasser, line 8:
```suggestion
```

PR #242, opened 2024-12-15
feat: add archive label and active status to ReviewModel

banesullivan:
For https://github.com/pyOpenSci/pyopensci.github.io/issues/320

This adds a new `active: bool = True` field to the `ReviewModel` class to track whether a package/review is archived or not. All reviews default to an `active=True` status and are only set to `active=False` if the `archived` label is applied to the original review.

In tandem with these changes, I've added a new "archived" label in the software-submissions repository. You can view this label and currently archived packages here: https://github.com/pyOpenSci/software-submission/issues?q=label%3Aarchived

If that specific label is applied to a review, then each time the pyosMeta `update-reviews` routine is run, it will discover that and set the `active` status to `False`

### DIFF
--- a/src/pyosmeta/models/github.py
+++ b/src/pyosmeta/models/github.py
@@ -17,9 +17,10 @@ datamodel-codegen --input issue_schema.json --input-file-type jsonschema --outpu
 from __future__ import annotations
 
 from datetime import datetime
+from enum import Enum
 from typing import Any, List, Literal, Optional, Union
 
-from pydantic import AnyUrl, BaseModel, ConfigDict, Field
+from pydantic import AnyUrl, BaseModel, ConfigDict, Field, model_validator
 
 
 class User(BaseModel):
@@ -61,6 +62,12 @@ class ClosedBy(User): ...
 class Owner(User): ...
 
 
+class LabelType(str, Enum):
+    ARCHIVED = "archived"
+    PYOS_APPROVED = "6/pyOS-approved"
+    JOSS_APPROVED = "9/joss-approved"
+
+
 class Labels(BaseModel):
     id: Optional[int] = None
     node_id: Optional[str] = None
@@ -69,6 +76,17 @@ class Labels(BaseModel):
     description: Optional[str] = None
     color: Optional[str] = None
     default: Optional[bool] = None
+    type: Optional[LabelType] = None
+
+    @model_validator(mode="before")
+    def parse_label_type(cls, data):
+        """Parse the label type from the name before validation."""
+        if "name" in data:
+            try:
+                data["type"] = LabelType(data["name"])
+            except ValueError:
+                pass
+        return data
 
 
 class Issue(BaseModel):
--- a/src/pyosmeta/models/github.py
+++ b/src/pyosmeta/models/github.py
@@ -77,10 +77,10 @@ class LabelType(str, Enum):
 
 
 class Labels(BaseModel):
+    name: str
     id: Optional[int] = None
     node_id: Optional[str] = None
     url: Optional[AnyUrl] = None
-    name: Optional[str] = None
     description: Optional[str] = None
     color: Optional[str] = None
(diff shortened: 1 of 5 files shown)


src/pyosmeta/models/github.py
```diff
@@ -69,6 +76,17 @@ class Labels(BaseModel):
     description: Optional[str] = None
     color: Optional[str] = None
     default: Optional[bool] = None
+    type: Optional[LabelType] = None
+
+    @model_validator(mode="before")
```
lwasser, line 81: It's been so long since i've dug into this code. let's flesh out the docstrings while this is still fresh. 

can you please add a second "paragraph" that better describes what we are actually parsing here and why as a reminder (for me an our future selves)?

src/pyosmeta/parse_issues.py
```diff
@@ -221,6 +221,25 @@ def _postprocess_meta(self, meta: dict, body: List[str]) -> dict:
 
         return meta
 
+    def _postprocess_labels(self, meta: dict) -> dict:
+        """
+        Handle specific labels that are model properties
```
lwasser, line 226: please be more specific with the docstrings. Essentially this package can be hard to "dig back into" so let's be specific about what this helper does since we don't have great docs for the package yet. 

Let's be specific about it's purpose in the docstring

src/pyosmeta/parse_issues.py
```diff
@@ -221,6 +221,25 @@ def _postprocess_meta(self, meta: dict, body: List[str]) -> dict:
...
+        """
+
+        def _is_archived(label: str) -> bool:
```
lwasser, line 229: this is a super explicit name but i think for the purposes of documenting process, over-documenting via docstrings is a good idea here. so let's add numpy docstrings with very clear descriptions of intent to each of these.

tests/integration/test_parse_issues.py
```diff
@@ -69,3 +69,20 @@ def test_parse_labels(issue_list, process_issues):
         issue.labels = labels
         review = process_issues.parse_issue(issue)
         assert review.labels == ["6/pyOS-approved", "another_label"]
+        assert review.active
+
+    # Now add an archive label
+    label_inst = Labels(
+        id=1196238794,
+        node_id="MDU6TGFiZWwxMTk2MjM4Nzk0",
```
lwasser, line 77: Codecov is noticing some test gaps - can you please ensure that we have full coverage via this pr?

lwasser (changes requested):
@banesullivan this pr works well! let's do the following
✅ please add more detailed docstrings to overdocument our process here. 
✅ please update the changelog in this pr
✅ make sure that we have full test coverage for this pr - there are a few gaps right now according to code cov output.

i'll then review again. essentially when this is good to go (it's really close!!) we can merge it and the website repo and the documentation updates to our peer review guide. 

Separately (new work in a document) i think we should work together on a hackmd that documents this entire process a bit better. (this is a separate task but one i don't want to forget to do!)

src/pyosmeta/parse_issues.py
banesullivan: I clarified the docstrings in b9868eb

src/pyosmeta/parse_issues.py
banesullivan: I added a docstring in b9868eb

tests/integration/test_parse_issues.py
banesullivan: Full coverage achieved ✅

src/pyosmeta/models/github.py
banesullivan: I clarified the docstrings in https://github.com/pyOpenSci/pyosMeta/commit/b9868eb69224b03a282d27cdc03c28e3ab34ee85

lwasser (approved):
Merging this and i'll add it to the release i started to create

PR #257, opened 2025-02-15
[hotfix] handle packages that lack a description on GitHub

banesullivan:
Hotfix for https://github.com/pyOpenSci/pyopensci.github.io/issues/560

The issue is that astrodata does not have a description on GitHub so the API returns `null` and our `GhMeta` Pydantic model requires this field be set to some string. Considering that repositories on GitHub don't have to have a description, I have changed the description field to be optional.

I also added a few other print statements to our routines to make it a little ore obvious what is going on. A few people had looked at the CI logs and thought Earthpy must be the issue since it was the last logged package. However that was from the previous processing step and the gh-metrics step was running across all packages without any indication.

This PR is meant as a hotfix to get CI back up and running but there are a few things I want to do to improve this which I will post in https://github.com/pyOpenSci/pyopensci.github.io/issues/560

### DIFF
--- a/src/pyosmeta/cli/process_reviews.py
+++ b/src/pyosmeta/cli/process_reviews.py
@@ -48,6 +48,7 @@ def main():
         print("-" * 20)
 
     # Update gh metrics via api for all packages
+    print("Getting GitHub metrics for all packages...")
     repo_endpoints = process_review.get_repo_endpoints(accepted_reviews)
     all_reviews = process_review.get_gh_metrics(
         repo_endpoints, accepted_reviews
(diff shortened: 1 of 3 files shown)


willingc (approved):
This looks good to me. Let's get this up and running. 

Someday in the future we may wish to migrate the print statements to logging but not important today.

pllim:
Thanks! Is there something else that needs to be done for the 4th package to show up at https://www.pyopensci.org/communities/astropy.html ? 🙏 🙇‍♀️

banesullivan:
I think we just need to issue a release for pyosmeta so that this new version is picked up by [this workflow](https://github.com/pyOpenSci/pyopensci.github.io/actions/workflows/update-contribs-reviews.yml) and then the site will be updated

banesullivan:
And that release to pypi is waiting on @lwasser's approval in https://github.com/pyOpenSci/pyosMeta/actions/runs/13348330244

Screenshot 2025-02-16 at 9 57 29 PM

Issue #191, opened 2024-07-12
[usability] ProcessIssues.parse_issues(issues) should have a print / count indicator so we know it's doing something


lwasser:
This is just a small feature description for the parse_issues method. right now the method works but it's hard to know if it's stalling because it's collecting data or because it is going to fail. i am on a bad internet connection and it's hard to tell right now!

A message that happens as it grabs each issue would be super helpful to know that the method is making progress. ✨ 

the print statement of log point would likely be [placed here](https://github.com/pyOpenSci/pyosMeta/blob/main/src/pyosmeta/parse_issues.py#L292)

PR #261, opened 2025-02-17
[usability] ProcessIssues.parse_issues(issues) should have a print /…count indicator so we know it's doing something #191

hariprakash619:
Added progress/count indicator in the def parse_issues() method.


lwasser:
Hi there! i am going to close this given we made a move to logging in #270 many thanks for your time in opening it!

PR #267, opened 2025-03-05
enh: use graphql to consolidate api calls

lwasser:
Next steps

- [x] Run on all packages
- [x] Consider a small reorg; put all gh metrics in the github module
- [x]  🐛 contrib count isn't correct.  (this is because this endpoint isn't available via graphQL. other options are mentionable and collaborators but those are different metrics.
- [x] Move get_gh_metrics to github module


lwasser:
bummer. if we move to graphql we can't get a contributor count. We could consider a separate workflow just for that metric in the future.

src/pyosmeta/cli/process_reviews.py
```diff
@@ -48,12 +48,10 @@ def main():
         print("-" * 20)
 
     # Update gh metrics via api for all packages
-    print("Getting GitHub metrics for all packages...")
-    repo_endpoints = process_review.get_repo_endpoints(accepted_reviews)
-    all_reviews = process_review.get_gh_metrics(
-        repo_endpoints, accepted_reviews
-    )
-
+    # BUG : contrib count isn't correct - great tables has some and is returning 0
```
lwasser, line 51: This isn't a bug, the endpoint isn't available via graphql.

lwasser: I've updated the PR to use rest just for that metric. it reduces the api calls by one per package vs two now.

src/pyosmeta/github_api.py
```diff
@@ -81,6 +83,10 @@ def get_token(self) -> str | None:
                 "Oops! A GITHUB_TOKEN environment variable wasn't found."
             )
 
+    # TODO: this property right now is tailored for the rest api grabbing
+    # pyopensci issues. Let's see after the refactor if we want to generalize it
+    # for other endpoints like contrib_count or if issues can move totally to
+    # graphql
     @property
```
lwasser, line 90:
```suggestion

    @property
```

lwasser:
pre-commit.ci autofix

lwasser:
Hey there, @banesullivan. I'm worried about merging this and breaking things in #270. Please have a look and let me know if we can merge and rebase or if that will cause problems. I'd also like to submit another pr that moves other parts of our workflow to graphql.

lwasser:
thank you @banesullivan !!!

PR #270, opened 2025-03-12
Consistent logging

banesullivan:
Make logging/debug statements consistently use a logger.

Note: we want to use `logging` over `warnings` as `warnings` will suppress a warning after the first time it is issued. When processing reviews, we want to see any warnings that are issued on a project by project basis

This also adds `tqdm` for progress monitoring


willingc (commented):
I love this approach @banesullivan. Will give a more thorough review when it is out of draft. Thank you!

src/pyosmeta/file_io.py
```diff
@@ -84,8 +86,8 @@ def open_yml_file(file_path: str) -> dict:
         with urllib.request.urlopen(file_path) as f:
             yaml = YAML(typ="safe", pure=True)
             return yaml.load(f)
-    except urllib.error.URLError as url_error:
-        print("Oops - can find the url", file_path, url_error)
+    except urllib.error.URLError:
+        logger.error(f"Oops - can find the url: {file_path}", exc_info=True)
```
lwasser, line 90:
```suggestion
        logger.error(f"Oops - I can't find the url: {file_path}", exc_info=True)
```

lwasser (approved):
thank you for cleaning this up @banesullivan let's merge.

PR #273, opened 2025-03-29
[Proposal] Autofix PRs with pre-commit CI

banesullivan:
Same as https://github.com/pyOpenSci/pyopensci.github.io/pull/603

This opts into automatically fixing PRs with pre-commit CI and removes `pre-commit/action` from the GitHub Actions workflows

- The  `pre-commit/action` GitHub Action is in maintenance only mode and they recommend swithing to pre-commit CI
- We previously used `pre-commit/action` to automatically reformat the YAML data files generated during the automated data update jobs.
- This PR will instead have those jobs automatically update the data, then pre-commit will automatically fix the formatting as a second commit.

This will remove the confusing error annotation we're seeing on the data update CI jobs since the pre-commit check technically fails since it reformats the YAML data files after they are generated

### DIFF
--- a/.github/workflows/test-update-contribs.yml
+++ b/.github/workflows/test-update-contribs.yml
@@ -30,9 +30,6 @@ jobs:
           update-reviews
           update-review-teams
 
-      - name: run precommit hooks
-        uses: pre-commit/action@v3.0.1
-        continue-on-error: true
       - name: Create Pull Request
         uses: peter-evans/create-pull-request@v7
         with:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -12,9 +12,11 @@
 # To run on all files: pre-commit run --all-files
 
 ci:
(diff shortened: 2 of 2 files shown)


willingc (approved):
Thanks @banesullivan

lwasser:
hey @banesullivan we intentionally don't use the autofix pr's for pre-commit CI because in our experience, they cause contributor problems. If the pr is fixed but we review a pr and contributors need to make changes, they need to know to force push to overwrite the PR OR to pull things down. 

i'm happy to autoimplement pre-commit ci for the contributor pr's but i think it's not a good idea for the rest of the website as we've run into too many issues with it. is there a way to only run pre-commit ci on specific pr's triggered by CI?

banesullivan:
precommit CI's docs have:

>(optional, default: true) whether to autofix pull requests. when disabled, comment / label "pre-commit.ci autofix" to a pull request to manually trigger auto-fixing.

So if we have the CI job that creates the PR also apply the `"pre-commit.ci autofix" label or comment that,  then we could just do it for these PRs

PR #287, opened 2025-05-20
Fix: remove sigstore step from publish build as it's duplicate

### LINKED ISSUE #286: Sigstore signing failure do to duplicate uploads
Today we successfully released a new version of pyosMeta to PyPI using a more secure // zizmor approved workflow. 

The sigstore signature failed because the distro was already uploaded in a previous step. 

It seems to be [uploaded here](https://github.com/pyOpenSci/pyosMeta/actions/runs/15123800032/job/42512002600#step:3:389) and then [again here in a new step (which fails)](https://github.com/pyOpenSci/pyosMeta/actions/runs/15123800032/job/42512002600#step:4:1)

Upon further reading, the sigstore action docs suggest that the other workflow ([PyPA's GitHub action](https://github.com/pypa/gh-action-pypi-publish) can handle the signature too. So there may not be a need to implement the sigstore action at all. (based on what i've read).

[I see this in the publish to PyPI build](https://github.com/pyOpenSci/pyosMeta/actions/runs/15123800032/job/42511886557#step:3:142). I'm just unclear of where and how the signature part is setup and if we can just remove that entire last sigstore step or not.

lwasser:
closes #286 

This fix removes a duplicate upload to sigstore. The PyPA action by default implements signing.

### DIFF
--- a/.github/workflows/publish-pypi.yml
+++ b/.github/workflows/publish-pypi.yml
@@ -72,4 +72,4 @@ jobs:
       - name: Publish package to PyPI
         # Only publish to PyPI on release
         if: github.event_name == 'release'
-        uses: pypa/gh-action-pypi-publish@76f52bc884231f62b9a034ebfe128415bbaabdfc
+        uses: pypa/gh-action-pypi-publish@76f52bc884231f62b9a034ebfe128415bbaabdfc # v1.12.4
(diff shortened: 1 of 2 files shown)


lwasser:
@all-contributors add @chenghlee for review

PR #290, opened 2025-06-11
[patch] support list of eic and editors

banesullivan:
I noticed that a number of reviews were failing to parse due to having multiple eic or editors. This fixes that


lwasser (approved):
Thank you @banesullivan !!

PR #292, opened 2025-07-12
Fix contributor count pagination - add per_page=100 to contributors A…

stevenrayhinojosa-gmail-com:
API call

- Resolves issue where contributor counts were capped at 30
- Now fetches all contributors using 100 pagination
- SunPy should show 239 contributors instead of 30
- We may need to debug at lines 236 -246 debug on github_api.py

### DIFF
--- a/src/pyosmeta/github_api.py
+++ b/src/pyosmeta/github_api.py
@@ -234,7 +234,8 @@ class GitHubAPI:
         logged, and the method returns None.
         """
         # https://api.github.com/repos/{owner}/{repo}/contributors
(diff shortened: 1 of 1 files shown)


lwasser:
@all-contributors please add @stevenrayhinojosa-gmail-com for code, review

Steven thank you for this! i think we can merge as it will enhance our contributor numbers but this bug still exists (i think) because i think we need to ensure the data from the api are paginated when contribs are over 100!

lwasser (approved):

lwasser:
Oops - this is throwing a different error. we can't merge yet until we troubleshoot.

lwasser (changes requested):
I'm unapproving while we troubleshoot the bug.

Issue #295, opened 2025-07-22
Bug: our contributor workflow is missing/skipping some contributors so there is likely a small bug


lwasser:
for instance, I added Steven as a contributor to our example package. and he is [not listed on our contributor page here](https://www.pyopensci.org/our-community/index.html). there are other contributors missing as well. something seems to be off with our workflow. 

https://github.com/pyOpenSci/pyosPackage?tab=readme-ov-file#contributors-

<img width="988" height="430" alt="Image" src="https://github.com/user-attachments/assets/2e7e346a-ac84-4b5e-ad9f-747b95dae6a3" />

Steven does have a dash in their name. so it could be that OR it could be that some of the newer repos aren't resolving properly

PR #296, opened 2025-07-25
fix: contrib types aligned with repos

lwasser:
This begins to address #295 
When going through the code, the contrib_type is based on the repo they contributed to. however in this case, the repo names were dated and in some cases missing so i've fixed that.


lwasser:
All tests are passing, and locally this catches new contributors. let's merge and see

PR #353, opened 2026-04-06
Doc: Add initial Sphinx docs structure

### LINKED ISSUE #181: Create docs for this package
I think it's time to create docs for this package. we probably want to use the pydata_sphinx theme given we've used this elsewhere but i'm open to using something simpler too like Furo or immaterial (mimicks mkdocs material). 

this issue is to just create the start of a documentation structure that we can publish online. 

the dstructure should have

* a developing / contributing section
* an about section.

we can start with a blank set of docs and slowly build them out pulling content from our readme file and the development docs when created in #180

Aniketsy:
fixes #181

### DIFF
--- a/docs/about/index.rst
+++ b/docs/about/index.rst
@@ -1,0 +1,20 @@
+About
+=====
+
+pyosMeta is the package used by pyOpenSci to update contributor metadata and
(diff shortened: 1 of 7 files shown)


willingc (approved):
Thanks @Aniketsy.

PR #368, opened 2026-07-11
fix: rename `_data` to match pyopensci.github.io structure

InessaPawson:
Update file paths to match the current `pyopensci.github.io` structure and restore compatibility with the contributor update workflow


lwasser:
@InessaPawson you know i forgot i pushed an update to the website - moving it to hugo. thank you for catching this. Hugo uses data/ not _data/ 

We will also need to fix this in our code throughout!! but i'll merge this now and we can fix the rest later too

lwasser (approved):
this is a good fix. i'm working on another pr to fix the package as well!!

lwasser:
@all-contributors please add @InessaPawson for bug, code

InessaPawson:
> We will also need to fix this in our code throughout!! 

I thought so, and I can take care of it. I just wanted to confirm that I was on the right track.

PR #377, opened 2026-08-05
feat: make paths in the package a constant defined in one place 

lwasser:
This just begins to make it easier to update paths if we need to in the future. I have not yet run tests on this. 

fixes #368


data/contributors.yml
lwasser, line 1: deleted intentionally so our workflow creates it for us again. i tested it locally and it works

PR #382, opened 2026-08-05
fix: permissions on the resuable workflow

lwasser:
This is a workflow permissions error that i think is fixed with read permissions


lwasser:
Cool - greenlight on my fork as long as everything else works we can merge this and it's fixed!!

PR #384, opened 2026-08-06
fix: make sure we never delete existing gh metrics + move to rest api

### LINKED ISSUE #380: bug: repository metrics are not updating AND are being deleted by pyosmeta.
we added a feature to detect packages that are not on github so we skip grabbing metrics for them. That part works but then it also skips over packages that are on github. It seems to lose the github metrics object data in the process so it deletes the data in the packages.yml file (metadata like stars contributors, etc). 

Update: i think there are some permission issues associated with pyosmeta. but what is problematic is it's DELETING existing metadata rather than not updating. so we need to fix tow things

1. fix the api calls to ensure we are grabbing what metrics are available publicly. (if anything now!)
2. Make sure it never deletes data entirely if data exist already for the package.

lwasser:
closes #380  

This specific pr fixes some bugs with collecting metadata for a package repository. The API enpoints and token requirements have changed. 
This 

* moves things to rest from graphql meaning we don't have to worry about elevated permissions
* Adds more specific failure points when we hit a 401, 403 so it's clear the points of failure.

IT adds Basic api defensive checks. 

It also ensures that metadata are never deleted by loading previous data from the ex packages.yml file first, just like we do with contributor data.

### DIFF
--- a/src/pyosmeta/cli/process_reviews.py
+++ b/src/pyosmeta/cli/process_reviews.py
@@ -104,6 +104,7 @@ def main():
     # Contrib count is only available via rest api
     logger.info("Getting GitHub metrics for all packages...")
     repo_paths = process_review.get_repo_paths(accepted_reviews)
+    # Grab existing gh metadata from the packages.yml file in the website repo
     existing_gh_meta = get_existing_gh_meta()
     all_reviews = github_api.get_metrics(
         repo_paths, accepted_reviews, existing_gh_meta
--- a/src/pyosmeta/cli/process_reviews.py
+++ b/src/pyosmeta/cli/process_reviews.py
@@ -24,9 +24,60 @@ To run at the CLI: parse_issue_metadata
 
 import pickle
 
+from pydantic import ValidationError
+
 from pyosmeta import ProcessIssues
+from pyosmeta.constants import PACKAGES_RAW_URL
+from pyosmeta.file_io import load_website_yml
 from pyosmeta.github_api import GitHubAPI
 from pyosmeta.logging import logger
+from pyosmeta.models.base import GhMeta
+
+
+def get_existing_gh_meta(url: str = PACKAGES_RAW_URL) -> dict[str, GhMeta]:
+    """Load the currently published packages.yml and pull out the
+    ``gh_meta`` block for each package, keyed by lowercased package name.
+
+    This is used as a fallback so that a failed GitHub API call doesn't
+    wipe out metrics we already collected in a previous run.
+
+    Parameters
+    ----------
+    url : str
+        URL to the live packages.yml file.
+
+    Returns
+    -------
+    dict
+        Mapping of lowercased package name to its last known ``GhMeta``.
+        Packages with no previously saved (or no longer valid) metrics are
+        omitted.
+    """
+    try:
+        existing_packages = load_website_yml("package_name", url)
+    except Exception:
+        logger.error(
+            "Couldn't load the existing packages.yml. Continuing without "
+            "a fallback for any GitHub metrics fetches that fail this run.",
+            exc_info=True,
+        )
+        return {}
+
+    existing_gh_meta = {}
+    for name, pkg in existing_packages.items():
+        if not pkg.get("gh_meta"):
+            continue
+        try:
+            existing_gh_meta[name] = GhMeta(**pkg["gh_meta"])
+        except ValidationError:
+            logger.error(
+                f"Existing gh_meta for {name} in the live packages.yml "
+                "doesn't match the current GhMeta model. Skipping it as a "
+                "fallback.",
+                exc_info=True,
+            )
+
+    return existing_gh_meta
 
 
 def main():
@@ -53,7 +104,10 @@ def main():
     # Contrib count is only available via rest api
     logger.info("Getting GitHub metrics for all packages...")
     repo_paths = process_review.get_repo_paths(accepted_reviews)
-    all_reviews = github_api.get_metrics(repo_paths, accepted_reviews)
+    existing_gh_meta = get_existing_gh_meta()
+    all_reviews = github_api.get_metrics(
+        repo_paths, accepted_reviews, existing_gh_meta
+    )
 
     with open("all_reviews.pickle", "wb") as f:
         pickle.dump(all_reviews, f)
--- a/src/pyosmeta/cli/process_reviews.py
+++ b/src/pyosmeta/cli/process_reviews.py
@@ -31,6 +31,7 @@ from pyosmeta.constants import PACKAGES_RAW_URL
 from pyosmeta.file_io import load_website_yml
 from pyosmeta.github_api import GitHubAPI
 from pyosmeta.logging import logger
+from pyosmeta.models import ReviewModel
 from pyosmeta.models.base import GhMeta
 
 
@@ -38,8 +39,8 @@ def get_existing_gh_meta(url: str = PACKAGES_RAW_URL) -> dict[str, GhMeta]:
     """Load the currently published packages.yml and pull out the
     ``gh_meta`` block for each package, keyed by lowercased package name.
 
-    This is used as a fallback so that a failed GitHub API call doesn't
-    wipe out metrics we already collected in a previous run.
+    Used by ``update_gh_meta`` as a backup when a fresh GitHub API fetch
+    leaves ``gh_meta`` empty. Does not itself apply metrics to reviews.
 
     Parameters
     ----------
@@ -53,6 +54,7 @@ def get_existing_gh_meta(url: str = PACKAGES_RAW_URL) -> dict[str, GhMeta]:
         Packages with no previously saved (or no longer valid) metrics are
         omitted.
     """
+    # TODO: fail fast if packages.yml cannot be loaded (separate follow-up).
     try:
         existing_packages = load_website_yml("package_name", url)
     except Exception:
@@ -80,6 +82,35 @@ def get_existing_gh_meta(url: str = PACKAGES_RAW_URL) -> dict[str, GhMeta]:
     return existing_gh_meta
 
 
+def update_gh_meta(
+    existing_gh_meta: dict[str, GhMeta],
+    reviews: dict[str, ReviewModel],
+) -> dict[str, ReviewModel]:
+    """Fill in missing gh_meta from previously published packages.yml.
+
+    Fresh API data (already on ``review.gh_meta``) wins. If a fetch left
+    ``gh_meta`` as None, reuse the last known ``GhMeta`` for that package.
+    """
+    for pkg_name, review in reviews.items():
+        if review.gh_meta is not None:
+            continue
+
+        previous_metadata = existing_gh_meta.get(pkg_name.lower())
+        if previous_metadata is not None:
+            logger.warning(
+                f"Couldn't refresh GitHub metrics for {pkg_name}. "
+                "Using the previously saved metrics from packages.yml."
+            )
+            review.gh_meta = previous_metadata
+        else:
+            logger.warning(
+                f"No GitHub metrics available for {pkg_name} "
+                "(none saved previously, and this fetch failed)."
+            )
+
+    return reviews
+
+
 def main():
     github_api = GitHubAPI(
         org="pyopensci",
@@ -104,11 +135,10 @@ def main():
     # Contrib count is only available via rest api
     logger.info("Getting GitHub metrics for all packages...")
     repo_paths = process_review.get_repo_paths(accepted_reviews)
-    # Grab existing gh metadata from the packages.yml file in the website repo
+    # Fetch first; gap-fill from packages.yml only where the API left gh_meta empty
     existing_gh_meta = get_existing_gh_meta()
-    all_reviews = github_api.get_metrics(
-        repo_paths, accepted_reviews, existing_gh_meta
-    )
+    all_reviews = github_api.get_metrics(repo_paths, accepted_reviews)
+    all_reviews = update_gh_meta(existing_gh_meta, all_reviews)
 
     with open("all_reviews.pickle", "wb") as f:
         pickle.dump(all_reviews, f)
--- a/src/pyosmeta/cli/process_reviews.py
+++ b/src/pyosmeta/cli/process_reviews.py
@@ -1,13 +1,13 @@
 """
-Script that parses metadata from and issue and adds it to a .yml file for the
+Script that parses metadata from an issue and adds it to a .yml file for the
 website. It also grabs some of the package metadata such as stars,
 last commit, etc.
 
 This script also checks:
-* That each packages documentation is both https compliant and resolves
+* That each package's documentation is both https compliant and resolves
 * If documentation link is broken it removes it.
 * NOTE: we may want to have the website note if docs are not available for a
-package and have a process to followup
+package and have a process to follow-up
 
 Output: packages.yml file containing a list of:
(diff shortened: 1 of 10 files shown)


src/pyosmeta/utils_clean.py
```diff
@@ -50,25 +50,33 @@ def get_clean_user(username: str) -> str:
 def clean_date(source_date: str | None) -> datetime | str:
     """Cleans up a date string to a consistent datetime format.
 
-    The source date string may have been manually entered as month-day-year format,
-    retrieved from GitHub as a timestamp, could be missing, or contain random text.
-    This utility validates the input and returns a consistent format.
+    The source date string may have been manually entered as month-day-year
```
lwasser, line 53: Essentially if we import existing data we don't have to clean the date because it's already nice and tidy.

src/pyosmeta/github_api.py
```diff
@@ -390,12 +516,18 @@ def get_repo_meta_github(
...
         -----
-        This method makes a GraphQL call to the GitHub API to retrieve metadata
-        about a pyos reviewed package repository.
+        REST (`_get_metrics_rest`) is the primary and only source used here
+        for repository metadata, so this works with a general token without
+        requiring org membership or elevated permissions. Contributor count
+        is fetched separately via `_get_contrib_count_rest`, since it isn't
+        part of the main repo metadata endpoint.
+
+        `_get_metrics_graphql` is kept temporarily as deprecated, non-critical
```
lwasser: Not removing the graphql until the end. essentially running into permission issues with it which is why i'm going back to rest calls

src/pyosmeta/cli/process_reviews.py
```diff
@@ -53,7 +104,11 @@ def main():
     # Contrib count is only available via rest api
     logger.info("Getting GitHub metrics for all packages...")
     repo_paths = process_review.get_repo_paths(accepted_reviews)
-    all_reviews = github_api.get_metrics(repo_paths, accepted_reviews)
+    # Grab existing gh metadata from the packages.yml file in the website repo
```
lwasser: Right now when gh metadata fail, we pull in 0 metadata for each repo and so the data get overwritten. Ideally we use the existing metadata rather than deleting it entirely.

src/pyosmeta/github_api.py
lwasser: this has now been removed it was the last step!

lwasser:
@willingc i also asked copilot to review as i'm curious what it will add to the pr. but i'd love a review from you if you have time.

lwasser:
rebasing this against main now - locally and squashing commits to simplify the process so the work is no longer isolated in scoped commits

src/pyosmeta/github_api.py
```diff
@@ -20,18 +20,62 @@
...
 
 from pyosmeta.models import ReviewModel
-from pyosmeta.models.base import RepositoryHost
+from pyosmeta.models.base import GhMeta, RepositoryHost
 
 from .logging import logger
 
 
+class GitHubAPIError(Exception):
+    """Raised for GitHub API errors that should stop the whole metrics run.
```
willingc:
```suggestion
    """Raised for GitHub API errors.
    
       The usual handling of these errors would be to stop the metrics run.
```

willingc: For clarity, it probably makes sense to define the errors in this docstring (401 and 403.

Handling and implementation (such as stopping the API call) should be documented where the action is being taken.

src/pyosmeta/github_api.py
```diff
@@ -20,18 +20,62 @@
...
+
 @dataclass
 class GitHubAPI:
     """
     A class that processes GitHub issues in our peer review process and returns
     metadata about each package.
     """
 
+    # Mapping for how GhMeta fields are populated.
+    # `contrib_count` is populated separately from the contributors REST endpoint.
```
willingc, line 46: Move this line somewhere else or into the docstring. It's a little confusing here.

src/pyosmeta/github_api.py
```diff
@@ -122,6 +166,26 @@ def api_endpoint(self) -> str:
 
         return f"{base_url}?{'&'.join(params)}"
 
+    @staticmethod
+    def _is_rate_limit_exhausted(response) -> bool:
+        """Return True if response headers show the primary rate limit is
+        fully used up, as opposed to a plain permission-denied 403."""
```
willingc:
```suggestion
        """Return True if rate limit has been reached.
        
        The header is checked for `X-RateLimit-Remaining` to indicate if the
        primary rate limit is
        fully used up, as opposed to a plain permission-denied 403."""
```

src/pyosmeta/github_api.py
```diff
@@ -122,6 +166,26 @@ def api_endpoint(self) -> str:
...
+        return response.headers.get("X-RateLimit-Remaining") == "0"
+
+    @staticmethod
+    def _format_rate_limit_reset(response) -> str:
+        """Format the X-RateLimit-Reset header (a Unix timestamp) as a
+        human-readable UTC time for clear error logs."""
```
willingc:
```suggestion
        """Format the X-RateLimit-Reset header timestamp to a
        human-readable UTC time."""
```

src/pyosmeta/github_api.py
```diff
@@ -199,25 +276,64 @@ def get_metrics(
             A dictionary mapping package names to their owner and repo-names.
         reviews : dict
             A dictionary containing review data.
+        existing_gh_meta : dict[str, GhMeta], Optional
+            A dictionary mapping lowercased package names to the last known
+            good ``GhMeta`` (e.g. loaded from the live packages.yml). Used as
```
willingc:
```suggestion
            good ``GhMeta`` name (e.g. loaded from the live packages.yml). Used as
```

src/pyosmeta/github_api.py
```diff
@@ -199,25 +276,64 @@ def get_metrics(
...
+            a fallback so a single failed API call doesn't erase previously
+            collected metrics.
 
         Returns:
         -------
         dict
             Updated review data with GitHub metrics.
         """
+        existing_gh_meta = existing_gh_meta or {}
+        stop_early = False
```
willingc:
```suggestion
        # if True, metrics run should be stopped
        stop_metrics_run = False
```

`stop_early` is a little imprecise.

src/pyosmeta/github_api.py
```diff
@@ -199,25 +276,64 @@ def get_metrics(
...
             with logging_redirect_tqdm():
                 review = reviews[pkg_name]
-                if review.repository_host == RepositoryHost.github:
-                    reviews[pkg_name].gh_meta = self.get_repo_meta_github(
-                        owner_repo
+                if review.repository_host != RepositoryHost.github:
+                    logger.warning(
+                        f"Unsupported repository host for {pkg_name}: {review.repository_host}"
+                    )
+                    continue
```
willingc, line 311: Do you want stop_run_early set to True here or just log a warning?

src/pyosmeta/github_api.py
```diff
@@ -199,25 +276,64 @@ def get_metrics(
...
+
+                previous_meta = existing_gh_meta.get(pkg_name.lower())
+
+                new_meta = None
```
willingc: A bit nit-picky but any reason not to be more explicit and use `previous_metadata` or `old_metadata` and `new_metadata`.

src/pyosmeta/github_api.py
```diff
@@ -199,25 +276,64 @@ def get_metrics(
...
+                if not stop_early:
```
willingc:
```suggestion
                if not stop_metrics_run:
```

willingc: This reads a bit easier to me without having to think too much.

willingc (approved):
Overall, it looks good. I made a few wording suggestions for maintainability but fine as is too.

src/pyosmeta/cli/process_reviews.py
```diff
@@ -24,9 +24,60 @@
...
+    Returns
+    -------
+    dict
+        Mapping of lowercased package name to its last known ``GhMeta``.
+        Packages with no previously saved (or no longer valid) metrics are
+        omitted.
+    """
+    try:
+        existing_packages = load_website_yml("package_name", url)
+    except Exception:
```
abadger, line 60: Disclaimer: if, in normal use, we sometimes run this script when there is no `packages.yml`, then this is probably not a correct analysis.

Do we want to stop the run on error here rather than continuing? I can see pros and cons to either approach:

* The reason for this PR is to prevent losing previous data if there is an error when retrieving data from the network. In this case, we have an error on the network (in normal use, a `packages.yml` should always already exist, right?) so we should fail and try again later?

* On the other hand, if we can retrieve all of the data from the REST API when this script runs without errors and we are only loading the old data to take care of cases where new information fails to load (off and on), then it seems fair to attempt to go on anyways.  At worst, this specific run will omit some relevant data. A subsequent run should restore it.

Maybe it boils down to whether we think it's less disruptive to users to have out-of-date data or data which flip-flops between omission and presence?

src/pyosmeta/cli/process_reviews.py
```diff
@@ -53,7 +104,11 @@ def main():
     # Contrib count is only available via rest api
     logger.info("Getting GitHub metrics for all packages...")
     repo_paths = process_review.get_repo_paths(accepted_reviews)
-    all_reviews = github_api.get_metrics(repo_paths, accepted_reviews)
+    # Grab existing gh metadata from the packages.yml file in the website repo
+    existing_gh_meta = get_existing_gh_meta()
+    all_reviews = github_api.get_metrics(
+        repo_paths, accepted_reviews, existing_gh_meta
+    )
```
abadger: I see that you had to enhance the cleaning of dates to account for pre-cleaned dates as well as the raw dates. I wonder if that's a sign we should be treating these as two datasets to merge rather than as data to validate the new data against?  Like this:

```suggestion
    existing_gh_meta = get_existing_gh_meta()
    new_gh_meta = github_api.get_metrics(
        repo_paths, accepted_reviews, existing_gh_meta
    )
    all_reviews = update_gh_meta(existing_gh_meta, new_gh_meta)
```

(We have to write `update_gh_meta()` in this case)

willingc:
I love seeing @abadger's review here. Great to have you lending your experience 💐

src/pyosmeta/cli/process_reviews.py
lwasser: @abadger this is actually a really good point AND we could consider a larger fail fast mechanism here (I think). For context, this workflow updates 2 data files in our [website repository ](https://github.com/pyOpenSci/pyopensci.github.io/tree/main/data). 

We are adding this element of "please update the existing data first" just in case we hit API errors, and github acn't be found and it tries to rewrite the data (which is what it was doing!). 

So if we cant find the yml file here, then we also can't update the files that this workflow intends to keep current. 

Should we instead add a fail fast mechanism to each entry point and fail the entire action if the files can't be found? 

Below is the live action that updates our website.

https://github.com/pyOpenSci/pyopensci.github.io/blob/main/.github/workflows/update-contribs-reviews.yml#L27

If so, perhaps we open a new issue and implement that separately. If i am not processing your point correctly, then pls let me know!

src/pyosmeta/cli/process_reviews.py
lwasser: ahhh yes, that is exactly what I ran into. I have to go back to the code to think about how that would work. Are you thinking that the `update_gh_meta()` function / method would have the date clean validation step (like we had before to parse the new data from the api)?

src/pyosmeta/github_api.py
lwasser: thank you!

src/pyosmeta/cli/process_reviews.py
lwasser: i wonder if we look for a file not found error or 404 here... thinking outloud but know it needs to be handled more cleanly!

src/pyosmeta/cli/process_reviews.py
```diff
@@ -53,7 +135,10 @@ def main():
     # Contrib count is only available via rest api
     logger.info("Getting GitHub metrics for all packages...")
     repo_paths = process_review.get_repo_paths(accepted_reviews)
+    # Fetch first; gap-fill from packages.yml only where the API left gh_meta empty
+    existing_gh_meta = get_existing_gh_meta()
     all_reviews = github_api.get_metrics(repo_paths, accepted_reviews)
+    all_reviews = update_gh_meta(existing_gh_meta, all_reviews)
```
lwasser, line 141: Added this method following @abadger review.

src/pyosmeta/github_api.py
lwasser: fixed!

lwasser:
@all-contributors please add @InessaPawson for review

lwasser:
@all-contributors please add @abadger for code, review

src/pyosmeta/cli/process_reviews.py
```diff
@@ -24,9 +24,91 @@
...
+        URL to the live packages.yml file.
+
+    Returns
+    -------
+    dict
+        Mapping of lowercased package name to its last known ``GhMeta``.
+        Packages with no previously saved (or no longer valid) metrics are
+        omitted.
+    """
+    # TODO: fail fast if packages.yml cannot be loaded (separate follow-up).
```
lwasser, line 57: I opened https://github.com/pyOpenSci/pyosMeta/issues/388 so we could think about fail fast comprehensively vs only fixing it here.

src/pyosmeta/github_api.py
```diff
@@ -159,33 +223,46 @@ def _get_response_rest(self, url: str) -> list[dict[str, Any]]:
...
-                raise exception
+            # Handle pagination & rate limiting
+            api_endpoint_url = response.links.get("next", {}).get("url")
+            self.handle_rate_limit(response)
 
         return results
 
     def get_metrics(
         self,
         endpoints: dict[dict[str, str]],
```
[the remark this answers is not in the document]
lwasser: This was actually a valid catch. i fixed it locally.

src/pyosmeta/github_api.py
lwasser: I think it's reasonable! i think because the package is called pyosMeta i've just always leaned into abbreviating! but i did go back and expand to be more expressive throughout! Good call.

src/pyosmeta/github_api.py
lwasser: oooh missed this comment. let me go back and have a look

src/pyosmeta/github_api.py
lwasser: fixed!!

src/pyosmeta/github_api.py
lwasser: thanks!!

src/pyosmeta/github_api.py
lwasser: ok yes we log a warning (could be a lower level info etc too) when we have a package hosted somewhere that we don't have api support for yet. Gitlab could come next but we only have a few on bitbucket, gitlab codeberg right now so it hasn't been a priority to support it yet. i think this is ok as is for now! good catch!!

lwasser:
Because this is still breaking our website build, i am going to merge this as is but i opened #388 to properly address yaml files and failing fast as a next step and we can open any other issues needed. I am going to merge, create a release and the builds should look better after that! and then we can finesse the code more

PR #393, opened 2026-08-25
feat: add get team members to workflow

lwasser:
This is a pr that begins to add the feature of reading team members to pyosmeta. right now i had to run through rest vs graph ql and landed on rest with a token scoped to this org, to read members. opening this in draft and will add tests and test a bit more + ci next.

### DIFF
--- a/data/contributors.yml
+++ b/data/contributors.yml
@@ -2,6 +2,8 @@
   github_username: lwasser
   github_image_id: 7649194
   title:
+    - Editor
+    - Peer Review Triage
     - Executive Director & Founder
     - Emeritus editor
   sort: 1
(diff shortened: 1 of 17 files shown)


data/contributors.yml
```diff
@@ -367,16 +369,14 @@
...
-    - Advisory Council
-    - Editor
+  title: Advisory Council
   sort:
   bio: Filipe Fernandes is a Physical oceanographer by training, turned
       research software engineer, and a software packager hobbyist.
   organization:
   date_added: '2019-10-17'
   deia_advisory: false
-  editorial_board: true
```
lwasser, line 379: So the idea is that if they are no longer on the editorial board, they are set to false automatically.

data/contributors.yml
```diff
@@ -558,14 +558,13 @@
...
   github_image_id: 3659681
   title:
-    - Emeritus Editor
   sort: 3
   bio: Researcher in Medical Informatics and Digital Health ⌚ 👩‍💻 🧬
   organization: '@hpi-dhc'
   date_added: '2021-08-24'
   deia_advisory: true
   editorial_board: false
-  emeritus_editor: true
```
lwasser, line 568: this however is incorrect as ariane is emeritus so this is a bug

PR #400, opened 2026-08-25
Fix:  emeritus editors not being updated correctly

lwasser:
This was just a small bug in the workflow where the teams were missing

### DIFF
--- a/data/manual-editorial-roster.yml
+++ b/data/manual-editorial-roster.yml
@@ -1,0 +1,27 @@
+# Hand-maintained editorial roster overrides.
+# --------------------------------------------
+# This file contains people who haven't joined our organization
+# but who do hold active or emeritus editorial roles.
+# This file is manually maintained and likely won't need to be
+# updated in the future as our onboarding is now better.
+# This file is merged by update-editorial-board after we fetch each team from
+# the github teams. Do not overwrite this file.
+
+# Here we define what flags to use for each person (what is true)
+# editor: true          → editorial-board.yml (active editor)
+# emeritus_editor: true → emeritus-editors.yml
+
+jbencook:
+  name: Ben Cook
+  note: "Not on GitHub teams; keep as emeritus editor"
+  emeritus_editor: true
+
+dhomeier:
+  name: Derek Homeier
+  note: "Not on GitHub teams; keep as active editor"
+  editor: true
+
+kellyrowland:
+  name: Kelly L. Rowland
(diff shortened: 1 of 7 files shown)


src/pyosmeta/constants.py
```diff
@@ -67,4 +67,7 @@
     "eic_team": "eic-team",
     "peer_review_lead": "peer-review-lead",
     "triage_team": "triage-team",
+    "emeritus_eic": "emeritus-editor-in-chief",
```
lwasser, line 75: Missing new team names!

src/pyosmeta/editorial.py
```diff
@@ -1,9 +1,8 @@
 """Build the editor roster from GitHub org team membership.
 
-GitHub teams are the source of truth for who is active vs emeritus.
-``emeritus_eic``, ``emeritus_peer_review_lead``, and ``emeritus_triage``
-are preserved from the existing website YAML because those roles are
-historical, not a current team.
+GitHub teams are the source of truth for active and emeritus specialty
+roles (``eic-team``, ``triage-team``, ``emeritus-editor-in-chief``,
+``emeritus-peer-review-lead``, ``emeritus-triage-team``, etc.).
```
lwasser:
```suggestion
``emeritus-peer-review-lead``, ``emeritus-triage-team``, "emeritus-editor"

.).
```

data/manual-editorial-roster.yml
```diff
@@ -0,0 +1,27 @@
+# Hand-maintained editorial roster overrides.
```
lwasser, line 1: this is clunky but there are a few people from our review history and present that just don't want to join our org. so here we manually curate the list since we can't add them to a team. if they ever join us, then we can remove from this list. or if someone new can't join the org, we can update the list.

src/pyosmeta/editorial.py
```diff
@@ -83,40 +92,71 @@ def build_roster(
...
-            emeritus_eic=bool(prev.get("emeritus_eic")),
-            emeritus_triage=bool(prev.get("emeritus_triage")),
+            emeritus_peer_review_lead=username in emeritus_pr_lead,
+            emeritus_eic=username in emeritus_eic,
+            emeritus_triage=username in emeritus_triage,
         )
     return roster
 
 
+def merge_manual_roster(
```
lwasser, line 126: this is a bit annoying but i'm not sure how else to curate manual additions.

